PR #8, opened 2024-05-07
chore(docker)[OK-35] Dockerize

Wajktor13:


docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '3'
+
+services:
+
+  frontend:
+    build: ./frontend
```
marcepanowyy, line 6: container_name: frontend_container

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    ports:
+      - "3000:3000"
+    depends_on:
+      - backend
+
+  backend:
+    build: ./backend
```
marcepanowyy, line 12: container_name: backend_container

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '3'
+
+services:
+
+  frontend:
+    build: ./frontend
+    ports:
+      - "3000:3000"
+    depends_on:
```
marcepanowyy: not necessary
(when db is up, then backend depends on db)

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+      - backend
+
+  backend:
+    build: ./backend
+    ports:
+      - "8080:8080"
```
marcepanowyy: I would suggest adding common network for frontend & backend (and later for db) to let them communicate with each other within the Docker cluster (otherwise its not possible)

    services: 
        front/backend: 
            networks:
                - my_network
        networks:
            my_network:
                driver: bridge

Wajktor13: Docker compose by default assigns all containers to the same network ([source](https://docs.docker.com/compose/networking/)). The default `driver` is also `bridge` ([source](https://docs.docker.com/network/drivers/)).

There is no need to explicitly declare that imo.

marcepanowyy (approved):

PR #13, opened 2024-06-01
feat(db)[OK-21] Add backend & db integration, base schemas

marcepanowyy:


backend/src/app.module.ts
```diff
@@ -1,10 +1,38 @@
...
+// const dbPort = process.env.DB_PORT || '30000';
 
 @Module({
-  imports: [],
-  controllers: [AppController],
-  providers: [AppService],
+  imports: [
+    MongooseModule.forRoot(
+      // `mongodb://${dbUser}:${dbPassword}@${dbHost}:${dbPort}/${dbName}`,
+      `mongodb://localhost/mongo_obelisk`,
```
Wajktor13, line 22: Why do we use this and not the commented line above?

backend/src/enums/user.type.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum UserTypeEnum {
```
Wajktor13: I'd change this to `UserAuthProvider` or something similar. `UserTypeEnum` is unambiguous for me.

backend/src/schemas/board.schema.ts
```diff
@@ -0,0 +1,56 @@
...
+    view: User[];
+    share: User[];
+  };
+
+  // JSON object
+  @Prop({
+    required: false,
+    type: [String],
+  })
+  slides: string[];
```
Wajktor13, line 47: I'm not sure here, but didn't we agree that we need a separate class (entity) `Slides`? Even if it might seem to be redundant for now, in the future we might want to store additional information for each slide.

backend/src/schemas/user.schema.ts
```diff
@@ -0,0 +1,71 @@
...
+    default: UserRoleEnum.USER,
+  })
+  userRole: UserRoleEnum;
+
+  @Prop({
+    required: false,
+    enum: UserTypeEnum,
+    default: UserTypeEnum.NORMAL,
+  })
+  userType: UserTypeEnum;
```
Wajktor13: As said previously, I'd consider more clear name, like `authProvider`.

backend/src/shared/validation.pipe.ts
Wajktor13, line 1: Is that file commented out to pass the linter?

backend/src/users/users.controller.spec.ts
Wajktor13, line 1: I know that the purpose of this test is to make CI pipeline pass, but I don't think we want such a bypass here. It doesn't bother me that much though if others agree to temporarily keep it. @Strattoss @Adam0s007

backend/src/users/users.controller.ts
```diff
@@ -0,0 +1,18 @@
+import { Body, Controller, Get, Post } from '@nestjs/common';
+import { CreateUserDto } from './users.dto';
+import { UsersService } from './users.service';
+
+@Controller('users')
+export class UsersController {
+  constructor(private usersService: UsersService) {}
+
+  @Post()
+  create(@Body() createUserDto: CreateUserDto) {
```
Wajktor13, line 10: I see some warning from backend pipeline here. Can't we use ValidationPipe you've created?

backend/src/users/users.dto.ts
```diff
@@ -0,0 +1,23 @@
...
+  // @Length(6, 30)
+  email: string;
+
+  @IsString()
+  @IsNotEmpty()
+  // @Length(8, 30)
+  // @Matches(/^(?=.*[A-Z].*[A-Z])(?=.*[!@#$%^&*]).{8,}$/, {
+  //   message:
+  //     'Password must have at least 8 characters, 2 uppercase letters, and 2 special characters.',
+  // })
```
Wajktor13, line 21: Why are those lines commented out?

Wajktor13 (changes requested):

Wajktor13:
General note. This PR is quite big and it covers many changes on different levels. It could easily be separated into several PRs. Let's try to keep PRs small. They ideally should cover one logical change if possible. This way it is not only easier to review such PRs, but also easier to develop the code.

@Strattoss @Adam0s007 applies to you as well.

backend/src/enums/user.type.enum.ts
```diff
@@ -0,0 +1,5 @@
...
+  NORMAL = 'normal',
```
Wajktor13: `INTERNAL = 'internal'` works better for me.

Wajktor13 (changes requested):

backend/src/app.module.ts
marcepanowyy: It will be used later

backend/src/schemas/board.schema.ts
marcepanowyy: I wasnt quite sure how to handle this. For now it may remain as it is. In the future I will make changes.

backend/src/shared/validation.pipe.ts
marcepanowyy: yeah, I couldnt find a proper way to resolve this issue

backend/src/users/users.controller.spec.ts
marcepanowyy: If no tests are present, CI throws errors

backend/src/users/users.controller.ts
marcepanowyy: The warning suggests using validation pipes individually for each endpoint. For now, it is used globally and I believe Linter doesnt see this.

backend/src/users/users.dto.ts
marcepanowyy: For testing purposes. I find it quite boring writing every time password that specifically matches the regex. You may waste time on that which is not quite what my intention is

backend/src/enums/user.type.enum.ts
marcepanowyy: agree on that

backend/src/users/users.controller.spec.ts
marcepanowyy: Besides, not only should linter `find `spec files, but also with `test`  suffix as well as shouldnt throw errors when no test are present...

backend/src/app.module.ts
Wajktor13: What is preventing us from using it now then?

backend/src/users/users.controller.spec.ts
Wajktor13: This is not the scope of this PR but ok.

I dont't think that linter is responsible for problems you described.

Look [here](https://github.com/OBELISK-TEAM/OBELISK/blame/c6ce95d4f11b8699853101ddff60499b5eedb70c/backend/package.json#L60). It was you who specified regex for test files (i.e. it was generated automatically).

Again, [here](https://github.com/OBELISK-TEAM/OBELISK/blame/c6ce95d4f11b8699853101ddff60499b5eedb70c/backend/package.json#L16) you can see that you declared how tests should be run (i.e. it was generated automatically). If we want the tests to pass even if there are no tests, this line should look like this: `"test": "jest --passWithNoTests"`.

I'll make Jira tasks for this and take it right away.

backend/src/users/users.dto.ts
Wajktor13: Yeah, but we should treat `develop` branch like production branch. I don't think we want it to be commented.

backend/src/users/users.controller.spec.ts
Wajktor13: I've created PRs for those changes.

Wajktor13: As the changes have been merged, the test should no longer be needed after updating this branch.

backend/src/app.module.ts
marcepanowyy: Locally there are no users pre-configured. If you want to connect locally to the db, you are supposed to start the MongoDB shell `mongo`, then switch to admin `use admin`, and create the admin user `db.createUser({
  user: "admin",
  pwd: "password",
  roles: [{ role: "userAdminAnyDatabase", db: "admin" }]
})` then, enable authentication (usually changes in the file `mongod.conf`) `security: authorization: enabled` and finally restart MongoDB and add `useNewUrlParser: true, useUnifiedTopology: true` to `Mongoose Module`.

Wajktor13: So 'later' means on production? Or when do we want to use it?
Won't Docker be helpful here?

marcepanowyy: Yeah, we can use it in production. I see no point when developing....

Wajktor13: Is there a point of leaving it commented out there?

Wajktor13 (approved):

PR #17, opened 2024-07-07
feat(slide schema)[OK-21] Add schema for slide & reorganize code

marcepanowyy:
Add schema for slide,
Reorganize code structure,
Add slide module,
Modify board schema


backend/src/enums/user.auth.provider.ts
```diff
@@ -1,5 +1,5 @@
 export enum UserAuthProvider {
-  INTERNAL = 'internal',
-  GOOGLE = 'google',
-  GITHUB = 'github',
+  INTERNAL = 1,
+  GOOGLE = 2,
+  GITHUB = 3,
```
Wajktor13, line 4: Whats the reason of this change?

backend/src/enums/user.role.ts
```diff
@@ -1,4 +1,10 @@
 export enum UserRole {
-  ADMIN = 'admin',
-  USER = 'user',
+  // can change roles, delete users, etc.
+  ADMIN = 100,
+
+  // can delete users, create/delete/modify boards, etc.
+  MEMBER = 10,
```
Wajktor13, line 6: I don't like this name at all, maybe `MODERATOR`?

Wajktor13 (approved):

Wajktor13 (changes requested):
.

backend/src/enums/user.role.ts
Wajktor13: I don't like this name at all, maybe `MODERATOR`?

Wajktor13 (changes requested):

backend/src/enums/user.role.ts
Wajktor13: I'm also not sure if we need this role

backend/src/enums/user.auth.provider.ts
marcepanowyy: imo its better to store numbers than strings in db

backend/src/enums/user.role.ts
marcepanowyy: i'll delete this role

Wajktor13 (approved):

PR #23, opened 2024-07-13
chore(docker-envs): [OK-61]: Add envs handling

Wajktor13:
After the merge, `.env` file will be required **at the root of the project** (where docker-compose.yml is located).  Format of the file:

```
# frontend
FRONTEND_HOST = 'localhost'
FRONTEND_PORT = '4000'

# backend
BACKEND_HOST = 'localhost'
BACKEND_PORT = '8080'
JWT_SECRET = 'secret'
JWT_EXPIRES_IN = '7d'

# database
DB_NAME = 'mongo_obelisk'
DB_USER = 'admin'
DB_PASSWORD = 'password'
DB_HOST = 'localhost'
DB_PORT = '27017'
```

Required envs have also been set in github actions envs.


marcepanowyy:
If you could only update the branch i would then try to fix the backend code so as it uses .env variable from the correct lvl file

marcepanowyy (approved):

PR #24, opened 2024-07-17
feat(layout)[OK-41] Add pages: login, signup, signup with email

Strattoss:


adbista:
All I can say is, good job! The Google icon will probably be changed to one from OK-78, but it can stay here for now. It's great that you added responsive design!

adbista (approved):

PR #26, opened 2024-07-18
chore(eslint)[OK-48] Add frontend eslint configuration

Strattoss:
The PR adds a more complex eslint configuration for the frontend. 

From now on eslint:

- shouts at you mercilessly if some imports are unused
- informs if a variable defined with `let` can be defined with `const` instead
- lets you know if you forget about some console.log or debug statement

and many more...

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,3 +1,53 @@
 {
-  "extends": "next/core-web-vitals"
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:react/recommended",
+    "plugin:react-hooks/recommended",
+    "plugin:jsx-a11y/recommended",
+    "plugin:import/errors",
+    "plugin:import/warnings",
+    "plugin:import/typescript",
+    "next/core-web-vitals"
+  ],
+  "plugins": [
+    "react",
+    "@typescript-eslint",
+    "jsx-a11y",
+    "import"
+  ],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "project": "./tsconfig.json"
+  },
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "off",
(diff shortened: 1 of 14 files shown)


frontend/.eslintrc.json
```diff
@@ -1,3 +1,27 @@
 {
-  "extends": "next/core-web-vitals"
-}
+  "plugins": [
+    "@typescript-eslint"
+  ],
+  "extends": [
+    "next/core-web-vitals",
+    "prettier",
+    "plugin:@typescript-eslint/recommended"
```
adbista: Is this the only plugin that needs to be added? I noticed you added more ESLint packages, so perhaps we should include additional plugins like these:

```suggestion
    "plugin:@typescript-eslint/recommended",
    "plugin:@jsx-a11y/recommended",
    "plugin:react-hooks/recommended",
    "plugin:react/recommended"
```
And so on.

frontend/.eslintrc.json
```diff
@@ -1,3 +1,27 @@
...
+        "*.tsx"
+      ],
+      "parserOptions": {
+        "project": [
+          "./tsconfig.json"
+        ]
+      }
+    }
+  ],
+  "rules": {
```
adbista, line 23: Maybe we should also disable rule with  no-explicit-any. I know it is the right thing to avoid using any, but i don't think at any cost we should always do this. sometimes we can got into troubles with that and i think we will have to even add ticket for resolving issues with that.
```suggestion
  "rules": {
  "@typescript-eslint/no-explicit-any": "off",
   "eslint-plugin-import/extensions": "off"
  }
}
```

adbista (changes requested):

frontend/.eslintrc.json
Strattoss: Yep, I agree

frontend/.eslintrc.json
Strattoss: I think you are right

adbista (approved):

PR #33, opened 2024-07-27
chore(docker)[OK-65] Dockerize mongodb, fix frontend Dockerfile

Wajktor13:
This PR introduces the following changes:

- dockerize mongodb
- set mongodb connection strings on backend
- fix frontend Dockerfile
- update `INSTALL.md` docs with info about running the app with `docker compose`

### DIFF
--- a/backend/src/app.module.ts
+++ b/backend/src/app.module.ts
@@ -1,5 +1,5 @@
 import { Module } from '@nestjs/common';
-import { MongooseModule } from '@nestjs/mongoose';
+import { MongooseModule, MongooseModuleFactoryOptions } from '@nestjs/mongoose';
 import { APP_FILTER, APP_INTERCEPTOR } from '@nestjs/core';
 import { HttpErrorFilter } from './shared/filters/http.error.filter';
 import { LoggingInterceptor } from './shared/interceptors/logging.interceptor';
@@ -10,16 +10,33 @@ import { ConfigModule, ConfigService } from '@nestjs/config';
 import { AuthModule } from './modules/auth/auth.module';
 import { PassportModule } from '@nestjs/passport';
 
+const DEFAULT_DB_HOST = 'localhost';
+
+function getMongoUri(
+  configService: ConfigService,
+): MongooseModuleFactoryOptions {
+  const dbName = configService.get<string>('DB_NAME');
+  const dbUser = configService.get<string>('DB_USER');
+  const dbPassword = configService.get<string>('DB_PASSWORD');
+  const dbHost = configService.get<string>('DB_HOST');
+  const dbPort = configService.get<string>('DB_PORT');
+
+  if (dbHost === DEFAULT_DB_HOST) {
+    return { uri: `mongodb://${DEFAULT_DB_HOST}/${dbName}` };
+  }
+
+  return {
+    uri: `mongodb://${dbUser}:${dbPassword}@${dbHost}:${dbPort}/`,
+    dbName: dbName,
+  };
+}
+
 @Module({
   imports: [
-    ConfigModule.forRoot({ isGlobal: true }),
+    ConfigModule.forRoot({ isGlobal: true, envFilePath: '../.env' }),
     MongooseModule.forRootAsync({
       imports: [ConfigModule],
-      useFactory: () => {
-        return {
-          uri: `mongodb://localhost/mongo_obelisk`,
-        };
-      },
+      useFactory: (configService: ConfigService) => getMongoUri(configService),
       inject: [ConfigService],
     }),
     PassportModule.register({ session: true }),
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,30 +1,51 @@
-version: '3'
-
 services:
 
   frontend:
-    build: ./frontend
+    build:
+      context: ./frontend
+      args:
+        - BACKEND_HOST
+        - BACKEND_PORT
     ports:
       - "${FRONTEND_PORT}:${FRONTEND_PORT}"
     container_name: obelisk_frontend_container
     environment:
-      - FRONTEND_HOST=${FRONTEND_HOST}
-      - FRONTEND_PORT=${FRONTEND_PORT}
-      - BACKEND_HOST=${BACKEND_HOST}
-      - BACKEND_PORT=${BACKEND_PORT}
+      - FRONTEND_HOST
+      - FRONTEND_PORT
 
   backend:
     build: ./backend
     ports:
       - "${BACKEND_PORT}:${BACKEND_PORT}"
     container_name: obelisk_backend_container
+    depends_on:
+      - mongo
     environment:
-      - BACKEND_HOST=${BACKEND_HOST}
-      - BACKEND_PORT=${BACKEND_PORT}
-      - DB_NAME=${DB_NAME}
-      - DB_USER=${DB_USER}
-      - DB_PASSWORD=${DB_PASSWORD}
-      - DB_HOST=${DB_HOST}
-      - DB_PORT=${DB_PORT}
-      - JWT_SECRET=${JWT_SECRET}
-      - JWT_EXPIRES_IN=${JWT_EXPIRES_IN}
+      - BACKEND_HOST
+      - BACKEND_PORT
+      - DB_NAME
+      - DB_USER
+      - DB_PASSWORD
+      - DB_HOST=mongo # must be set to mongo
+      - DB_PORT
+      - JWT_SECRET
+      - JWT_EXPIRES_IN
+      - GOOGLE_EMAIL
+      - GOOGLE_PASSWORD
+      - GOOGLE_CLIENT_ID
+      - GOOGLE_CLIENT_SECRET
+      - GOOGLE_CALLBACK_URL
+      - SESSION_SECRET
+      - COOKIE_SESSION_MAX_AGE
+      - THROTTLE_TTL
+      - THROTTLE_LIMIT
+
+  mongo:
+    image: mongo:latest
+    ports:
+      - "27017:27017" # must be set to 27017
+    container_name: obelisk_mongodb_container
+    environment:
+      - MONGO_INITDB_DATABASE=${DB_NAME}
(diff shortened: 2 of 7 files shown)


docker-compose.yml
```diff
@@ -28,3 +26,15 @@ services:
       - DB_PORT=${DB_PORT}
```
marcepanowyy: image

The endpoints are not mapped in nestJS application, I cannot perform simple registering

image

I would suggest testing simple endpoints to see whether docker files are working or not. I see no relation between backend & mongo inside docker. The only thing is that the db is initialized first (depend_on property in dockero-compose file), before backend.

marcepanowyy (changes requested):

docker-compose.yml
Wajktor13: Of course changes have been tested and it works like a charm for me 😉.

Creating a user:
image

Get on users:
image

Nest logs:
image


I think it proves that everything works and you must be doing something wrong.


Following the [Installation Guide](https://github.com/OBELISK-TEAM/OBELISK/blob/OK-65/docs/INSTALL.md) should fix the problems. I suppose you don't have correct envs set.

Wajktor13:
@Adam0s007 @Strattoss could you check if it works as expected for you? Please follow the [Installation Guide](https://github.com/OBELISK-TEAM/OBELISK/blob/OK-65/docs/INSTALL.md).

docker-compose.yml
marcepanowyy: my mistake, old images were used

marcepanowyy:
After deleteting deprecated images and creating new ones, the issues have appeared:

backend: 
image

frontend: 
image
image

Moreover, not only me having problem with frontend container... @Strattoss

Wajktor13:
> After deleteting deprecated images and creating new ones, the issues have appeared:
> 
> backend: image
> 
> frontend: image image
> 
> Moreover, not only me having problem with frontend container... @Strattoss

How are you running it?

marcepanowyy:
docker-compose up --build -d

Wajktor13:
> docker-compose up --build -d

Could you try setting `FRONTEND_HOST` env to `0.0.0.0`? It should make frontend work. Please confirm.

When it comes to backend, I cannot reproduce the issue on Ubuntu. Will try clean install on Windows.

Strattoss:
Well, regarding me, it seems like I also have some problems but they are not the same as @marcepanowyy 's. Just so you know, I also work on Windows. After carefully following the INSTALL.md instructions (I use `docker compose build` and `docker compose up`), I can conclude that:
- Backend works just fine for me. Nestjs successfully connects to mongodb. Postman requests work correctly. 
obraz
- Frontend with docker works correctly, but only after changing `FRONTEND_HOST` to `0.0.0.0` 🪄. Otherwise it's impossible to connect to it just like @marcepanowyy described.
- Frontend doesn't work locally. It's impossible to run it from Powershell, because Powershell (or rather next) doesn't recognize the "${FRONTEND_PORT}" syntax (or frankly any other I can think of). Have you tried running it locally @Wajktor13 ? If it works for you, I guess it could be because I use Windows and you work on Linux.
obraz

Wajktor13:
> Well, regarding me, it seems like I also have some problems but they are not the same as @marcepanowyy 's. Just so you know, I also work on Windows. After carefully following the INSTALL.md instructions (I use `docker compose build` and `docker compose up`), I can conclude that:
> 
>     * Backend works just fine for me. Nestjs successfully connects to mongodb. Postman requests work correctly.
>       obraz
> 
>     * Frontend with docker works correctly, but only after changing `FRONTEND_HOST` to `0.0.0.0` 🪄. Otherwise it's impossible to connect to it just like @marcepanowyy described.
> 
>     * Frontend doesn't work locally. It's impossible to run it from Powershell, because Powershell (or rather next) doesn't recognize the "${FRONTEND_PORT}" syntax (or frankly any other I can think of). Have you tried running it locally @Wajktor13 ? If it works for you, I guess it could be because I use Windows and you work on Linux.
>       obraz

The `${FRONTEND_PORT}` syntax I've used is Linux - specific indeed (didn't know that). I am working on some workaround.

It works in the docker image, because it is based on Alpine Linux.

Strattoss:
> > Well, regarding me, it seems like I also have some problems but they are not the same as @marcepanowyy 's. Just so you know, I also work on Windows. After carefully following the INSTALL.md instructions (I use `docker compose build` and `docker compose up`), I can conclude that:
> > ```
> > * Backend works just fine for me. Nestjs successfully connects to mongodb. Postman requests work correctly.
> >   obraz
> > 
> > * Frontend with docker works correctly, but only after changing `FRONTEND_HOST` to `0.0.0.0` 🪄. Otherwise it's impossible to connect to it just like @marcepanowyy described.
> > 
> > * Frontend doesn't work locally. It's impossible to run it from Powershell, because Powershell (or rather next) doesn't recognize the "${FRONTEND_PORT}" syntax (or frankly any other I can think of). Have you tried running it locally @Wajktor13 ? If it works for you, I guess it could be because I use Windows and you work on Linux.
> >   obraz
> > ```
> 
> The `${FRONTEND_PORT}` syntax I've used is Linux - specific indeed (didn't know that). I am working on some workaround.
> 
> It works in the docker image, because it is based on Alpine Linux.

I wish you luck then.

The workaround I have found so far is to change scripts in such a way, that `npm run dev` runs `node dev-server.js`, and the `dev-server.js` script reads `.env` file manually and only then it runs `next dev -p ${FRONTEND_PORT}`. You can find it [here](https://stackoverflow.com/questions/60147499/how-to-set-port-in-next-js) in the less popular answer.

Ultimately we can simply state that next will always run on the default 3000 port, even while using docker (although we could simply configure docker in such a way, that is associates next's default 3000 port and exposes it as some other port, like 4000).

backend/src/app.module.ts
```diff
@@ -10,14 +10,27 @@ import { ConfigModule, ConfigService } from '@nestjs/config';
...
+const DEFAULT_DB_HOST = 'localhost'
+
 @Module({
   imports: [
-    ConfigModule.forRoot({ isGlobal: true }),
+    ConfigModule.forRoot({ isGlobal: true, envFilePath: '../.env' }),
     MongooseModule.forRootAsync({
       imports: [ConfigModule],
-      useFactory: () => {
+      useFactory: (configService: ConfigService) => {
```
marcepanowyy: I would suggest adding a method for getting the uri

marcepanowyy (commented):
when runinng docker compose build:
image

marcepanowyy:
Seems like there is still a problem, @Strattoss @Adam0s007 can you run the frontend?

image

Wajktor13:
> Seems like there is still a problem, @Strattoss @Adam0s007 can you run the frontend?
> 
> image

Use `0.0.0.0` instead of `localhost` as `FRONTEND_HOST`

Strattoss (approved):
Checked and tested. All works fine for me. Nextjs can be run locally, it also correctly reads variables from `.env` in the project root directory. The same things hold true for the docker container.

docs/INSTALL.md
```diff
@@ -7,20 +7,45 @@ This document contains information crucial for using and developing the project.
 An `.env` file with the following content is required at the root of the project (where `docker-compose.yml` is located): 
 
 ```
-# frontend
-FRONTEND_HOST = 'localhost'
-FRONTEND_PORT = '4000'
+# frontend - for docker only
+FRONTEND_HOST=0.0.0.0
+FRONTEND_PORT=4000
```
marcepanowyy: not 5000?

docs/INSTALL.md
```diff
@@ -7,20 +7,45 @@ This document contains information crucial for using and developing the project.
...
 
 # backend
-BACKEND_HOST = 'localhost'
-BACKEND_PORT = '8080'
```
marcepanowyy, line 16: 4000?

docs/INSTALL.md
```diff
@@ -7,20 +7,45 @@ This document contains information crucial for using and developing the project.
...
-JWT_SECRET = 'secret'
-JWT_EXPIRES_IN = '7d'
+BACKEND_HOST=localhost
+BACKEND_PORT=8080
+JWT_SECRET=secret
+JWT_EXPIRES_IN=7d
+GOOGLE_EMAIL=mail@mail.com
```
marcepanowyy: google email & password are redundant (it was for me only to check the Google Cloud Console)

docs/INSTALL.md
```diff
@@ -7,20 +7,45 @@ This document contains information crucial for using and developing the project.
...
+DB_PORT=27017 # must be set to 27017
+```
+## Running with docker compose
+
+First, install `docker` and `docker compose`.
+
+Run the following commands at the root of the project (where `docker-compose.yml` is located):
+
+```
+docker compose build
```
marcepanowyy: No need to run docker compose build when running for the first time - images will be build automatically. However when we are not sure whether we previously created the images we can simply run replacing two commands with one:
```
docker-compose up --build -d
``` 
--build = always rebuiling, -d = work i detach mode (run in the background)

marcepanowyy (approved):
Besides my comments, everything works fine

docs/INSTALL.md
Wajktor13: That's just an example like this wole entry in `INSTALL.md`. You can set whatever you want. Why `5000` exactly?

docs/INSTALL.md
Wajktor13: Like above.

docs/INSTALL.md
Wajktor13: Yeah, I know all that.  I just thought it would be better to include more general version in docs.

Besides, `docker-compose` (aka Docker Compose v1) has been deprecated for over a year now ([link](https://github.com/actions/runner-images/issues/9557)) so it should be `docker compose up --build -d` following the Docker Compose v2 syntax.

adbista (approved):

docs/INSTALL.md
marcepanowyy: 4000 is the default

PR #42, opened 2024-08-07
feat(Auth)[OK-57] Implementation and Refinement of Internal and Google Authentication

adbista:
### Summary of Changes

- **Integrated Google Login**: Implemented logic for user authentication via Google accounts.
- **Enhanced Internal Authentication**: Improved security with cookie-based login and registration endpoints.
- **Added Core Components**: Introduced middleware, `AuthContext`, and `MessageReceiver` for better state and message management.

#### `useAuthForm`
- Replaced separate `useState` hooks (like `isLoading`, `error`, etc.) with a single reducer to manage related states in the authentication form. This logic is wrapped in a custom hook for better readability and maintenance.

#### `useAuth`
- Provides a context with methods for making API requests without handling errors directly; these must be managed where the functions are invoked.
- **Note**: Although there's a `try-catch` block around these functions, the `catch` block only converts and rethrows the error.
- Manages messages by setting the appropriate message based on the success or failure of the API requests.
- Returns authentication functions along with `getAuthMessage` (for reading the API request status) and `clearAuthMessage` for imperative message management.

#### `handleApiError`
- Converts API errors into a standardized array of error messages and throws the error as a string array for consistent error handling.

#### `useHandleAuth`
- Utilizes the states from `useAuthForm` effectively, setting different parts of the reducer according to the current action (e.g., updating `isLoading` during API requests, setting `error` in the catch block, etc.).
- Invokes authentication-related functions such as `login` and `signup` from `AuthContext`, executing the appropriate function based on the `AuthAction` type and handling errors accordingly.
- This hook can be accessed in components and returns specialized handlers like `login: handleAuth(AuthAction.LOGIN)`, making it easy to use in component-level authentication actions.
- In summary, this hook serves as a wrapper around `useAuth` functions, simplifying their use in components.

#### `MessageReceiver`
- Accesses `useAuth` primarily for message management. It provides a UI for displaying these messages and includes functionality to remove notifications as needed.

#### `useGoogleAuth`
- Acts as a wrapper around `loginGoogleUser` from `useAuth`. It handles the creation of a new window and state management for Google authentication.

### Summary 

- **`useAuthForm`**: A primitive wrapper around `useReducer` for form state management.
- **`useAuth`**: A context providing functions for API requests and message management.
- **`handleApiError`**: Converts API errors into a standardized format.
- **`useHandleAuth`**: Combines and manages hooks from `useAuthForm` and `useAuth`, wrapping their functionality for easier use in components.
- **`MessageReceiver`**: Manages and displays messages from the `useAuth` context.
- **`useGoogleAuth`**: Acts as a wrapper around `loginGoogleUser` from useAuth.

Potential Enhancements:
- **`General Message Management`**: In the future, we can create a more generalized message management system by extracting the message handling logic from `useAuth` into a separate provider. This would allow for more flexible and reusable message handling across different parts of the application.


marcepanowyy (approved):

frontend/next.config.mjs
```diff
@@ -1,5 +1,29 @@
...
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+    env: {
+        BACKEND_HOST: process.env.BACKEND_HOST,
+        BACKEND_PORT: process.env.BACKEND_PORT,
+        GOOGLE_CLIENT_ID: process.env.GOOGLE_CLIENT_ID,
+        GOOGLE_CLIENT_SECRET: process.env.GOOGLE_CLIENT_SECRET,
+        GOOGLE_CALLBACK_URL: process.env.GOOGLE_CALLBACK_URL,
+    },
+    async headers() {
```
marcepanowyy, line 17: Unecessary i think

frontend/package.json
```diff
@@ -21,15 +22,23 @@
     "@radix-ui/react-slot": "^1.0.2",
     "@radix-ui/react-switch": "^1.0.3",
     "@radix-ui/react-toggle": "^1.0.3",
+    "@types/jsonwebtoken": "^9.0.6",
+    "@types/jwt-decode": "^3.1.0",
+    "axios": "^1.7.3",
```
marcepanowyy, line 27: Are you using axios? Props for using inbuild fetch function. But when not using axios anymore it should not be in the package json. (ofc other unused packages too)

https://stackoverflow.com/questions/22675725/how-to-find-unused-packages-in-package-json

Btw @Wajktor13 it could be another, nice one checker

marcepanowyy (changes requested):

marcepanowyy (dismissed):

marcepanowyy:
image

Locally works perfect, however, when using docker an error occurs

marcepanowyy:
https://stackoverflow.com/questions/74165121/next-js-fetch-request-gives-error-typeerror-fetch-failed

frontend/src/app/api/auth/google/route.ts
```diff
@@ -0,0 +1,26 @@
+import { NextRequest, NextResponse } from "next/server";
+import { setTokenCookie } from "@/lib/authApi";
+
+export async function POST(req: NextRequest) {
+  const { state } = await req.json();
+
+  const response = await fetch(
```
marcepanowyy: Shouldnt response be wrapped up in try-catch block? What if hitting the domain that does not exist?

frontend/src/app/api/auth/google/route.ts
```diff
@@ -0,0 +1,26 @@
...
+    },
+  );
+
+  if (response.ok) {
+    const { accessToken } = await response.json();
+    setTokenCookie(accessToken);
+    return NextResponse.json({ message: "Login successful" });
+  } else {
+    const error = await response.json();
+    return NextResponse.json(error, { status: response.status });
```
marcepanowyy: Argument type {status: number} is not assignable to parameter type ResponseInit | undefined

marcepanowyy:
@Wajktor13 could you take a look at the problem with the docker? Downgrading the node version to v18 does not work.
When trying to downgrade next version, sth bad happens and cant even compile it locally. 

https://stackoverflow.com/questions/74165121/next-js-fetch-request-gives-error-typeerror-fetch-failed 

@Adam0s007 could you confirm the error is related to docker and giving you the same error?

Cheers :)

> image
> 
> Locally works perfect, however, when using docker an error occurs

frontend/src/app/api/auth/google/route.ts
adbista: You are right, i am changing that now

frontend/src/app/api/auth/google/route.ts
adbista: When did you receive this kind of error?

marcepanowyy: In webstorm IDE (although the code compiles)

frontend/src/app/api/auth/google/route.ts
marcepanowyy: And maybe handling also 400/401/403/404?

adbista:
> @Wajktor13 could you take a look at the problem with the docker? Downgrading the node version to v18 does not work. When trying to downgrade next version, sth bad happens and cant even compile it locally.
> 
> https://stackoverflow.com/questions/74165121/next-js-fetch-request-gives-error-typeerror-fetch-failed
> 
> @Adam0s007 could you confirm the error is related to docker and giving you the same error?
> 
> Cheers :)
> 
> > image
> > Locally works perfect, however, when using docker an error occurs

yes

frontend/src/app/api/auth/google/route.ts
adbista: Message is going from the backend, i don't define it, i don't think this is necessary

frontend/src/app/api/auth/google/route.ts
adbista: Could you please provide more details about this error? Specifically, when did you encounter it? Which endpoint were you trying to access, what data did you send, and was Docker running at the time?

Adam0s007:
> @Wajktor13 could you take a look at the problem with the docker? Downgrading the node version to v18 does not work. When trying to downgrade next version, sth bad happens and cant even compile it locally.
> 
> https://stackoverflow.com/questions/74165121/next-js-fetch-request-gives-error-typeerror-fetch-failed
> 
> @Adam0s007 could you confirm the error is related to docker and giving you the same error?
> 
> Cheers :)
> 
> > image
> > Locally works perfect, however, when using docker an error occurs

frontend/next.config.mjs
```diff
@@ -1,12 +1,31 @@
...
 const nextConfig = {
     env: {
-      BACKEND_HOST: process.env.BACKEND_HOST,
-      BACKEND_PORT: process.env.BACKEND_PORT
+
+        BACKEND_HOST: process.env.BACKEND_HOST,
+        BACKEND_PORT: process.env.BACKEND_PORT,
+        GOOGLE_CLIENT_ID: process.env.GOOGLE_CLIENT_ID,
+        GOOGLE_CLIENT_SECRET: process.env.GOOGLE_CLIENT_SECRET,
+        GOOGLE_CALLBACK_URL: process.env.GOOGLE_CALLBACK_URL,
```
Wajktor13, line 15: Those envs are not passed by docker. `docker-compose.yml` and  `Dockerfile` should be updated.

Wajktor13 (changes requested):

Wajktor13 (approved):

marcepanowyy (approved):

frontend/src/hooks/auth/useGoogleAuth.ts
```diff
@@ -4,6 +4,12 @@ const getGoogleLoginUrl = () => {
   const state = crypto.randomUUID();
   const googleAuthEndpoint = "https://accounts.google.com/o/oauth2/v2/auth";
 
+  console.log(
+    "process.env.GOOGLE_CALLBACK_URL",
+    process.env.GOOGLE_CALLBACK_URL,
+  );
+  console.log("process.env.GOOGLE_CLIENT_ID", process.env.GOOGLE_CLIENT_ID);
```
Wajktor13: I'd remove those console logs, we don't want to leak those envs.

Wajktor13 (changes requested):

docker-compose.yml
```diff
@@ -4,14 +4,19 @@ services:
...
+        - GOOGLE_CLIENT_SECRET
+        - GOOGLE_CALLBACK_URL
     ports:
       - "${FRONTEND_PORT}:${FRONTEND_PORT}"
     container_name: obelisk_frontend_container
     environment:
       - FRONTEND_HOST
       - FRONTEND_PORT
+      - BACKEND_HOST=backend
+      - BACKEND_PORT
```
Wajktor13: Those envs are useless here.

Wajktor13 (changes requested):

docker-compose.yml
adbista: Okay, you are right, now it works for me without them.

Wajktor13 (approved):

marcepanowyy:
works fine with docker

frontend/src/lib/auth-api.ts
```diff
@@ -0,0 +1,16 @@
+export const apiRequest = async (endpoint: string, method: string, body?: any) => {
```
Strattoss: I'm not sure where we use it but it seems like it's a function so perhaps we could change it into a verb? Something like `performApiRequest` or `makeApiRequest` or something else, whatever you find suitable.

Strattoss (approved):

PR #43, opened 2024-08-09
style[OK-90] Reformat the code using prettier

marcepanowyy:
Depcheck has identified a missing dependency for express, even though this package is built into NestJS and doesn't need to be explicitly included in the project. Additionally, it flagged @nestjs/schematics as unused, which is actually necessary for scaffolding and generating new modules, although it’s not directly used in the application code, as well as ts-loader & tsconfig-paths (for resolving paths) & source-map-support (for tracking stack error).

Depcheck also detects prettier packages as unused which is false positive imo - they are needed.

So I did not modified these dependencies (packages)

image


docker-compose.yml
```diff
@@ -30,15 +30,9 @@ services:
...
       - JWT_EXPIRES_IN
-      - GOOGLE_EMAIL
-      - GOOGLE_PASSWORD
       - GOOGLE_CLIENT_ID
       - GOOGLE_CLIENT_SECRET
       - GOOGLE_CALLBACK_URL
-      - SESSION_SECRET
-      - COOKIE_SESSION_MAX_AGE
-      - THROTTLE_TTL
-      - THROTTLE_LIMIT
```
Wajktor13, line 41: I think we should also update `docs/INSTALL.md` to make it consistent.

Wajktor13 (approved):

PR #45, opened 2024-08-10
feat(slide-object)[OK-98] Add DTO, export complex types

Wajktor13:
I can't find reliable source of information about common form of serialized slide-object. Does anybody have it by any chance?


backend/src/schemas/slide-object.schema.ts
```diff
@@ -176,10 +179,10 @@ export class SlideObject {
   skewY: number;
 
   @Prop({
-    type: [[String, Number, Number]],
+    type: Array<SlideObjectPath>,
```
marcepanowyy: We need to decide on a single convention - either 
```
Array<sth>
```
or
```
sth[]
```
Second option is shorter and more consise imo and used in many places in the project

backend/src/shared/interfaces/SlideObjectFilter.ts
```diff
@@ -0,0 +1,4 @@
+export class SlideObjectFilter {
```
marcepanowyy: Shouldnt be there an interface? There is possibility to use class but for rather for dto with using decorators from class-validator: 
```
export class CreateUserDto {
  @IsEmail()
  @IsNotEmpty()
  @Length(6, 30)
  email: string;
```

Wajktor13: I get this error when I use an interface:

image

Wajktor13: I get this error when I use an interface:

image

backend/src/schemas/slide-object.schema.ts
Wajktor13: Changed to `[]` where I could, but in `@Prop` it must be `Array`:

image

marcepanowyy:
```
  @Prop({
    type: [
      {
        type: MongooseSchema.Types.Mixed,
        properties: {
          type: {
            type: String,
            required: true,
          },
          value: {
            type: Number,
            required: true,
          },
        },
      },
    ],
  })
  filters?: SlideObjectFilter[];
  ```
  with interface
  ```
  export interface SlideObjectFilter {
  type: string;
  value: number;
}

  ```

Wajktor13: i don't like this solution, because if something changes in the interface, you have to fix it in two spots.

Wajktor13: Also if we want to apply it to `SlideObjectTextStyles` it would get very complicated

marcepanowyy: properties are optional, you dont need to use them

Wajktor13: so just `[ { type: MongooseSchema.Types.Mixed}]` for all of the 'complex` types?

marcepanowyy: For now its enough imo

Wajktor13: Done

marcepanowyy (approved):

PR #46, opened 2024-08-11
feat(slide-objects)[OK-99] Add module, service and controller; fix DTO and schema

Wajktor13:
This PR adds module, service and controller for SlideObject.

I also decided to make all `Fabric.js`-specific fields in scheme optional, because its unconsistent which of them are really required (if any of them is).

The controller has been tested with all SlideObject example values from [the jira ticket](https://pawg.atlassian.net/jira/software/c/projects/OK/boards/2?selectedIssue=OK-84).

### DIFF
--- a/backend/src/modules/auth/users/users.service.ts
+++ b/backend/src/modules/auth/users/users.service.ts
@@ -5,6 +5,7 @@ import { CreateUserDto } from './users.dto';
 import { User, UserDocument } from '../../../schemas/user.schema';
 import { Board } from '../../../schemas/board.schema';
 import { UserAuthProvider } from '../../../enums/user.auth.provider';
+import { SlideObject } from 'src/schemas/slide-object.schema';
 
 @Injectable()
 export class UsersService {
@@ -68,5 +69,19 @@ export class UsersService {
       await existingUser.save();
     }
     return existingUser;
+  }
+
+  async addSlideObject(
+    userId: string,
+    slideObject: SlideObject,
+  ): Promise<void> {
+    const updatedUser = await this.userModel
+      .findByIdAndUpdate(
+        userId,
+        { $push: { slideObjects: slideObject } },
+        { new: true },
+      )
+      .exec();
+    if (!updatedUser) throw new HttpException('User not found', 404);
   }
 }
--- a/backend/src/modules/slide-objects/slide-objects.controller.ts
+++ b/backend/src/modules/slide-objects/slide-objects.controller.ts
@@ -1,0 +1,62 @@
+import {
+  Controller,
+  Get,
+  Post,
+  Put,
+  Delete,
+  Param,
+  Body,
+  HttpException,
+  HttpStatus,
+} from '@nestjs/common';
+import { SlideObjectsService } from './slide-objects.service';
+import {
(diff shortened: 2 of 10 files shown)


backend/src/modules/slide-objects/slide-objects.service.ts
```diff
@@ -0,0 +1,75 @@
...
+  constructor(
+    @InjectModel(SlideObject.name)
+    private readonly slideObjectModel: Model<SlideObject>,
+    private readonly userService: UsersService,
+    private readonly slideService: SlidesService,
+  ) {}
+
+  async findAll(): Promise<SlideObjectDocument[]> {
+    return this.slideObjectModel.find().exec();
+  }
```
Wajktor13, line 26: Just a note here. This is just for test purposes and will never be needed on production. Endpoints like this should be deleted / disabled on production, especially that it is quite heavy.

backend/src/modules/auth/users/users.service.ts
```diff
@@ -69,4 +70,18 @@ export class UsersService {
     }
     return existingUser;
   }
+
+  async addSlideObject(
+    userId: string,
+    slideObject: SlideObject,
```
marcepanowyy, line 76: You could have added Partial<SlideObject> and then no need to write @Optional for all the fields i guess

backend/src/modules/slide-objects/slide-objects.controller.ts
```diff
@@ -0,0 +1,91 @@
...
+  CreateSlideObjectDto,
+  UpdateSlideObjectDto,
+} from './slide-objects.dto';
+import { SlideObjectDocument } from 'src/schemas/slide-object.schema';
+
+@Controller('slide-objects')
+export class SlideObjectsController {
+  constructor(private readonly slideObjectsService: SlideObjectsService) {}
+
+  private handleError(error: unknown): never {
```
marcepanowyy: https://stackoverflow.com/questions/76900620/exception-handling-in-spring-controller-or-service-layer

image

I have also read that controllers should be a simple pass through and not be responsible for handling errors using try-catch block. Every request goes through http.error.filter which is already implemented (src/shared/filters/http.error.filter)

backend/src/modules/slide-objects/slide-objects.dto.ts
```diff
@@ -19,138 +19,138 @@ export class CreateSlideObjectDto {
...
   @IsString()
-  @IsNotEmpty()
-  version: string;
+  @IsOptional()
+  version?: string;
 
   @IsString()
-  @IsNotEmpty()
-  originX: string;
+  @IsOptional()
```
marcepanowyy: No need for @IsOptional when using Partial<T>

backend/src/modules/slide-objects/slide-objects.controller.ts
marcepanowyy: https://docs.nestjs.com/exception-filters

Wajktor13: whatever, did it your way.

marcepanowyy (approved):

PR #50, opened 2024-08-16
feat(board)[OK-59] Implement board eraser

Strattoss:
The eraser has been implemented thanks to a published fabricjs [eraser mixin](https://github.com/fabricjs/fabric.js/blob/master/src/mixins/eraser_brush.mixin.ts). There is the EraserBrush description:

```
  /**
   * EraserBrush class
   * Supports selective erasing meaning that only erasable objects are affected by the eraser brush.
   * Supports **inverted** erasing meaning that the brush can "undo" erasing.
   *
   * In order to support selective erasing, the brush clips the entire canvas
   * and then draws all non-erasable objects over the erased path using a pattern brush so to speak (masking).
   * If brush is **inverted** there is no need to clip canvas. The brush draws all erasable objects without their eraser.
   * This achieves the desired effect of seeming to erase or unerase only erasable objects.
   * After erasing is done the created path is added to all intersected objects' `eraser` property.
   *
   * In order to update the EraserBrush call `preparePattern`.
   * It may come in handy when canvas changes during erasing (i.e animations) and you want the eraser to reflect the changes.
   *
   * @tutorial {@link http://fabricjs.com/erasing}
   * @class fabric.EraserBrush
   * @extends fabric.PencilBrush
   * @memberof fabric
   */
```

Because the Eraser is a mixin, it isn't by default in the [npm fabric package](https://www.npmjs.com/package/fabric). That's why it has to be added into the application in the `"postinstall "` in the `package.json` file. Instruction on how to do it was found [here](https://www.npmjs.com/package/fabric-with-erasing).

Other problematic thing is the fact that the EraserBrush class is not in the `@types/fabric` package so the TS is very unhappy, because it thinks that we create and use an instance of a class that hasn't been defined anywhere. That's why we have to write `// @ts-ignore` everywhere we use it.

Edit: Scaling of erased or partialy erased object works correctly now.

### DIFF
--- a/frontend/.depcheckrc
+++ b/frontend/.depcheckrc
@@ -1,0 +1,2 @@
+ignores: ["uglify-js"]
+skip-missing: true
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "frontend",
       "version": "0.1.0",
+      "hasInstallScript": true,
       "dependencies": {
         "@radix-ui/react-alert-dialog": "^1.0.5",
         "@radix-ui/react-avatar": "^1.0.4",
@@ -37,6 +38,7 @@
         "jwt-decode": "^4.0.0",
         "lucide-react": "^0.378.0",
         "next": "14.2.3",
+        "postcss": "^8",
         "react": "^18",
         "react-dom": "^18",
         "tailwind-merge": "^2.3.0",
@@ -52,7 +54,8 @@
         "eslint-plugin-import": "^2.29.1",
         "eslint-plugin-jsx-a11y": "^6.9.0",
         "eslint-plugin-react": "^7.35.0",
-        "eslint-plugin-react-hooks": "^4.6.2"
+        "eslint-plugin-react-hooks": "^4.6.2",
+        "uglify-js": "3.19.2"
       }
     },
     "node_modules/@aashutoshrathi/word-wrap": {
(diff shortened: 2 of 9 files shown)


Strattoss:
How it actually works in action:

https://github.com/user-attachments/assets/21d4d52e-8f64-4728-b906-0f4ef3b6971b

adbista (changes requested):
There is an issue with the selection and drawing modes. When we add a new object to the canvas, the selection mode icon is no longer active. To address this, we could adjust our code by returning **isDrawingMode** in `useMenuData`. In `SidebarItem`, try modifying the `className` on the button like this:

```jsx
<button
  className={`flex items-center text-left text-gray-600 hover:bg-gray-200 p-2 rounded cursor-pointer ${
    activeItem === item.name ||
    (isDrawingMode && item.name === MenuAction.DrawingMode) ||
    (!isDrawingMode && item.name === MenuAction.SelectionMode)
      ? "bg-gray-200 text-black"
      : ""
  }`}
>
```

Consider adding a prop for `isDrawingMode` in `SidebarItem`, and then passing it from `BoardSidebar` in `Sidebar.tsx`. Finally, make sure to pass it through in `page.tsx` within the `board` folder. Additionally, when we select the eraser mode, the drawing mode becomes active as well. You can skip handling this for now. During the refactoring process, I'll change the isDrawingMode boolean type to something more descriptive, like canvasMode state, using an enum type with values such as
`draw`, `erase`, and `select`.

adbista (approved):

Strattoss:
Fix has been introduced.
As you've noticed, when a user chooses the eraser mode, the pencil icon will also be highlighted as if it was active.
Leaving it for now.

PR #52, opened 2024-08-18
feat(dark theme)[OK-64] Implement dark theme

Strattoss:
The PR contains:
- Synchronizing color values between Figma and Nextjs.
- Refactoring the application in such a way that a dark theme is possible.
- Adding the `next-themes` package.
- Preparing a button to toggle between light and dark themes.
- Making the app logo and the google icon work correctly with the dark theme


frontend/src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const ThemeProvider: FC<{children: React.ReactNode}> = ({children}) => {
+    const [darkTheme, setDarkTheme] = useState(() => {
+        if (typeof window !== 'undefined') {
+            const theme = localStorage.getItem('theme');
+          return (theme && theme === 'dark') ? true: false;
+        }
+        return false;
+      });
+
+    useEffect(() => {
```
adbista: Is it necessary? We are already doing this when initializing `useState`

frontend/src/app/Providers.tsx
```diff
@@ -2,11 +2,14 @@
 import React from "react";
 import { AuthProvider } from "@/contexts/AuthContext";
 import MessageReceiver from "@/app/MessageReceiver";
+import { ThemeProvider } from "@/contexts/ThemeContext";
 
 const Providers = ({ children }: { children: React.ReactNode }) => {
   return (
     <AuthProvider>
-      <MessageReceiver>{children}</MessageReceiver>
+      <ThemeProvider>
```
adbista: With the current approach, when navigating to a subpage like `/board` or `/user-boards`, the page initially loads with the light theme, and only after a moment does it switch to the dark theme if that's what we've chosen. This flicker between themes doesn't seem ideal.

I came across a tutorial on implementing dark mode in Next.js without this flicker. It might be useful for you to check it out:  
[Next.js Dark Mode with No Flicker + Tailwind CSS](https://www.youtube.com/watch?v=7zqI4qMDdg8).

adbista (changes requested):

frontend/src/contexts/ThemeContext.tsx
Strattoss: Good call. It used to be necessary but no longer is

frontend/src/app/Providers.tsx
Strattoss: Works like a charm!

adbista (approved):

PR #55, opened 2024-08-21
refactor(Frontend)[OK-89] Refactor & format code on frontend side

adbista:
The main changes include:

1. **MenuDataContext**: Introduced a custom hook `useMenuActions` to manage menu list actions.
2. **Toolbar**: Created a `Control` folder to better organize components within the Toolbar. Split the logic across multiple files to enhance maintainability.
3. **Board**: Refactored `fabricCanvasUtils` into multiple context-specific files. Moved interfaces, reducers, and types into dedicated board-related folders. Added comments for better code clarity.
4. **Hooks**: Removed unused custom hooks to clean up the codebase.
5. **UndoRedoContext**: Added a new `UndoRedoContext` for handling undo and redo functionality, improved `FileContext`, and introduced a `KeydownListenerWrapper`.
7. **General Refactoring**: Removed unused imports, fixed types and parameters, converted specific values to the correct type, and updated certain strings to use enums.
9. **devDependencies**: Moved some packages to `devDependencies` to optimize production build size.
10. **Drawing Mode**: Refactored related functions for improved functionality.
11. **Canvas Mode**: Added a function `isActiveItem` and adjusted mode handling to use enums more effectively.


frontend/src/components/board/HorizontalMenu.tsx
```diff
@@ -1,37 +1,31 @@
...
-    }
+const BoardHorizontalMenu: FC<HorizontalMenuProps> = ({ boardName, groupId }) => {
+  const {
+    state: { activeItem, selectedObjectStyles },
+  } = useCanvas();
+  const { menuList } = useMenuData();
+  const menuItems = menuList.find((group) => group.groupId === groupId);
+
+  const handleClick = (action?: () => void) => {
+    //console.log("wykonuje sie teraz");
```
Strattoss: I think we can remove that

frontend/src/components/board/Toolbar/Controls/ObjectSpecificControls.tsx
Strattoss, line 1: I'm wondering if it could be possible to implement this control in a different way.
We would look at the canvas object's properties. If the object has a `color` property, then we render a color control. If the object has a `width` property, we render a width control. And so on.

Just to be clear, I don't think it should be a part of this branch. If anything, it could be a separate Jira task. That whole idea is just my proposition

frontend/src/enums/CanvasMode.ts
```diff
@@ -0,0 +1,5 @@
+export enum CanvasMode {
+  Selection = "selection-mode",
+  SimpleDrawing = "drawing-mode",
+  Eraser = "eraser-mode",
+}
```
Strattoss: I'm not sure why we decided to name enums with small letters, but I would much rather prefer capital letter. So it would be 
```suggestion
export enum CanvasMode {
  SELECTION = "selection-mode",
  SIMPLE_DRAWING = "drawing-mode",
  ERASER = "eraser-mode",
}
```

It's the same case for all the enums

frontend/src/hooks/board/useMenuActions.ts
```diff
@@ -0,0 +1,182 @@
...
+      return;
+    }
+    setCanvasMode(CanvasMode.Selection);
+  },
+  [MenuActions.ChangeColor]: () => {
+    // No-op: handled by color picker UI
+  },
+  [MenuActions.ChangeSize]: () => {
+    // No-op: handled by size picker UI
+  },
```
Strattoss: I'm thinking if we should throw an error here so that we know that we should not use these functions.

frontend/src/hooks/board/useMenuActions.ts
```diff
@@ -0,0 +1,182 @@
...
+      return;
+    }
+    exportToPDF(canvas);
+  },
+  [MenuActions.Undo]: () => {
+    // No-op: handled directly in MenuDataContext
+  },
+  [MenuActions.Redo]: () => {
+    // No-op: handled directly in MenuDataContext
+  },
```
Strattoss: Same thing here

frontend/src/interfaces/canvas-context.ts
```diff
@@ -0,0 +1,17 @@
+import { CanvasReducerState } from "@/types/CanvasReducer";
+import { MutableRefObject } from "react";
+import { CanvasMode } from "@/enums/CanvasMode";
+
+export interface ICanvasContext {
```
Strattoss: I understand that we couldn't have named it `CanvasContext` (or at least we shouldn't have), but I think that if we decide to name one of the interfaces using the `I`prefix, then we should be consistent and use this naming convention for all of our interfaces.
It's one or the other. Otherwise the names will be confusing

frontend/src/utils/board/undoRedoUtils.ts
Strattoss, line 1: This file is empty. Is it intended?

frontend/src/utils/handleApiError.ts
Strattoss, line 1: This whole file should be deleted. It's not used any more.

frontend/src/utils/board/canvasUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+    // scale the erased parts
+    if (obj.eraser) {
+      obj.eraser.scaleX *= initialObjScaleX;
+      obj.eraser.scaleY *= initialObjScaleY;
+    }
+  }
+  obj.setCoords();
+};
+
+//todo: make use of this zoom function
```
Strattoss, line 84: I think we can make a separate task out of it. Or even better - we could write a user story, which would precisely determine what we actually want to achieve.

Strattoss (changes requested):
An astounding work. I hope that now it will be easier to write code and add functionalities to the board :)
As soon as you resolve my minor change requests, I will approve the PR

frontend/src/hooks/board/useMenuActions.ts
adbista: okay, good point, i will do it

frontend/src/hooks/board/useMenuActions.ts
adbista: Okay

frontend/src/interfaces/canvas-context.ts
adbista: Good point, probably i will add I prefix for every interface

frontend/src/utils/board/undoRedoUtils.ts
adbista: ymm, right, redundant file

frontend/src/utils/handleApiError.ts
adbista: yes right

frontend/src/utils/board/canvasUtils.ts
adbista: yes, i just wanted  to write it somewhere, we have lots of work now, so this feature won't be implemented in a week, but yea, we can create a task and move to backlog

frontend/src/components/board/Toolbar/Controls/ObjectSpecificControls.tsx
adbista: Yes, i am creating task for that

Strattoss (approved):

PR #57, opened 2024-08-22
feat(toast notifications)[OK-96] Implement general message management

Strattoss:
* Remove out previous custom MessageReceiver.
* Add out-of-the-box shadcn's solution: Sonner and Toasts.
* Fix the issue with Google failed authorization attempt. Formerly it was throwing an uncaught error.


frontend/src/contexts/toastAuthorizationResult.tsx
```diff
@@ -0,0 +1,37 @@
+import { ToastTypes } from "@/enums/ToastType";
+import React from "react";
+import { toast } from "sonner";
+
+export const toastAuthorizationResult = (
+  toastType: ToastTypes,
+  toastTitle: string,
```
adbista: Just a thought—why not make this field optional? If `toastType` and `toastTitle` are often the same, we could allow `toastType` to serve as the `toastTitle` when `toastTitle` isn’t defined.

frontend/src/app/Providers.tsx
```diff
@@ -1,14 +1,24 @@
...
+
 import React from "react";
 import { AuthProvider } from "@/contexts/AuthContext";
-import MessageReceiver from "@/app/MessageReceiver";
 import { ThemeProvider } from "next-themes";
+import { Toaster } from "@/components/ui/toaster";
+import { ToasterProps } from "sonner";
 
 const Providers = ({ children }: { children: React.ReactNode }) => {
+  const toastsConfig: ToasterProps = {
```
adbista: Why not move this logic to a dedicated wrapper for the Toaster? Since it’s only related to toast notifications, this area should be reserved for more generic code related to all (or at least two) providers. Consider adding a `<ToasterWrapper />` component, where you can define the `toastsConfig` and then return `<Toaster {...toastsConfig} />`.

adbista (changes requested):
image

As you can see, there is a large error message. The problem is that this error creates a poor user experience because the toast notification only appears for a few seconds, leaving the user without enough time to read the entire message. Consider displaying the toast for a longer duration for the auth/signup/email endpoint (perhaps 30 seconds) or even making it permanent so the user can manually close it. For other endpoints, the current duration can remain as those errors are brief.

Also since we will no longer use ErrorList, you can simply delete it.

frontend/src/app/Providers.tsx
Strattoss: Good call

frontend/src/contexts/toastAuthorizationResult.tsx
Strattoss: Done

adbista (approved):

PR #59, opened 2024-08-25
feat(auth)[OK-92] Allow internal register when email registered by Google

Wajktor13:


backend/src/modules/users/users.service.ts
```diff
@@ -24,6 +24,24 @@ export class UsersService {
     return createdUser.save();
   }
 
+  async update(
+    userId: string,
+    updateUserDto: UpdateUserDto,
+  ): Promise<UserDocument> {
+    // have to do it this way to trigger password hashing, findByIdAndUpdate won't work
```
marcepanowyy: Try adding in user.schema

```
UserSchema.pre('findByIdAndUpdate', function (next) {
  const update = this.getUpdate();
  if (update.password) {
    update.password = hashSync(update.password, 10);
  }
  next();
});
```

backend/src/modules/users/users.dto.ts
```diff
@@ -1,10 +1,13 @@
...
+  IsEnum,
   IsNotEmpty,
+  IsOptional,
   IsString,
   Length,
   Matches,
 } from 'class-validator';
+import { UserAuthProvider } from 'src/enums/user.auth.provider';
 
 export class CreateUserDto {
```
marcepanowyy, line 12: i think _class-transformer_ and _class-validators_ dont automatically validate objects when passing them from service to service

backend/src/modules/auth/auth.service.ts
```diff
@@ -62,10 +63,35 @@ export class AuthService {
...
+
+    return this.generateToken(user);
+  }
+
+  private async handleRegisterWhenEmailExists(
+    createUserDto: CreateUserDto,
+  ): Promise<UserDocument> {
+    const user = await this.usersService.findOneByEmail(createUserDto.email);
+
+    if (user.userAuthProvider != UserAuthProvider.GOOGLE) {
```
marcepanowyy: I would put here
```
if (user.userAuthProvider === UserAuthProvider.INTERNAL)
```
User could have different auth provider in the future

backend/src/modules/auth/auth.service.ts
```diff
@@ -62,10 +63,35 @@ export class AuthService {
...
-    return this.generateToken(newUser);
+    }
+
+    const updateUserDto = {
+      email: createUserDto.email,
+      password: createUserDto.password,
+      userAuthProvider: UserAuthProvider.INTERNAL_AND_EXTERNAL,
+    } as UpdateUserDto;
+
+    await this.usersService.update(user.id as string, updateUserDto);
```
marcepanowyy: i think there is **_id** property fields intead of **id**

marcepanowyy (changes requested):

backend/src/modules/users/users.dto.ts
Wajktor13: Not sure to what you are reffering specifically to, but we create `UpdateUserDto` based on received `CreateUserDto`, so no further validation is needed.

backend/src/schemas/user.schema.ts
```diff
@@ -72,3 +73,11 @@ UserSchema.pre('save', function (next) {
   this.password = hashSync(this.password, 10);
   next();
 });
+
+UserSchema.pre('findOneAndUpdate', function (next) {
+  const update = this.getUpdate() as UpdateQuery<User>;
+  if (update && typeof update.password == 'string') {
```
marcepanowyy: I think it should be:
```
update && update.password && typeof update.password == 'string'
```
since we want to hash when **password** is changed (what if updating user other properties beside password - then the old password would be hashed again i think)

marcepanowyy (changes requested):
I would change only one thing...

marcepanowyy (approved):

PR #63, opened 2024-08-25
feat(Zoom)[OK-111] Create Zoom functionality 

adbista:
## Add ZoomUIContext

### DIFF
--- a/frontend/src/contexts/CanvasContext.tsx
+++ b/frontend/src/contexts/CanvasContext.tsx
@@ -11,13 +11,15 @@ import {
   toggleDrawingMode,
   updateDimensions,
 } from "@/utils/board/canvasUtils";
+import { ZoomOptions } from "@/enums/ZoomOptions";
+import { useZoom } from "./ZoomUIContext";
 
 const CanvasContext = createContext<ICanvasContext | undefined>(undefined);
 
 export const CanvasProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
   const [state, dispatch] = useReducer(canvasReducer, initialState);
   const canvasRef = useRef<HTMLCanvasElement | null>(null);
-
+  const { handleZoom } = useZoom();
   useEffect(() => {
     const newCanvas = initializeCanvas({ current: canvasRef.current });
     dispatch({ type: CanvasReducerAction.SET_CANVAS, canvas: newCanvas });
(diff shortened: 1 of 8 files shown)


frontend/src/contexts/CanvasContext.tsx
```diff
@@ -53,6 +55,35 @@ export const CanvasProvider: React.FC<{ children: React.ReactNode }> = ({ childr
       updateDimensions(obj);
     };
 
+    const handleMouseWheel = (opt: any) => {
+      const evt = opt.e;
+      if (newCanvas) {
```
Strattoss: Can we please use early return?
I even have a poem for it!

One indentation less,
Makes my brain not digress
```suggestion
      if (!newCanvas) {
            return;
      };
```

frontend/src/contexts/ZoomUIContext.tsx
```diff
@@ -0,0 +1,47 @@
+"use client";
+import React, { createContext, useContext, useState, useRef, useEffect } from "react";
+import { ZoomOptions } from "@/enums/ZoomOptions";
+
+interface ZoomContextProps {
```
Strattoss: I thought we wanted to name the interfaces of contexts the same way as the contexts themselves. If so, it should be `ZoomContext`.
Or we could move this interface into the `interfaces` directory and import it as `IZoomContext`.
If, however, we want to name context interfaces with the `*Props` suffix, then we should be consistent and use this convention for **every** context.

frontend/src/contexts/CanvasContext.tsx
```diff
@@ -11,13 +11,15 @@ import {
...
+import { useZoom } from "./ZoomUIContext";
 
 const CanvasContext = createContext<ICanvasContext | undefined>(undefined);
 
 export const CanvasProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
   const [state, dispatch] = useReducer(canvasReducer, initialState);
   const canvasRef = useRef<HTMLCanvasElement | null>(null);
-
+  const { handleZoom } = useZoom();
   useEffect(() => {
```
Strattoss, line 23: I think that this `useEffect` should have the `handleZoom` function in its dependecies. At least that's what the linter is anxious about.

Strattoss (changes requested):
What a nifty feature!

Strattoss (approved):

PR #66, opened 2024-08-28
docs(boards)[OK-118] Complete Swagger docs

Wajktor13:


backend/src/shared/interfaces/response-objects/BoardResponseObject.ts
```diff
@@ -1,4 +1,4 @@
-export interface BoardResponseObject {
+export class BoardResponseObject {
```
marcepanowyy: I think there is no need to use class here. Classes are useful in combination with transformer and validator package to determine the correct dto is passed. Simple interface should be good enough. 

https://www.reddit.com/r/reactnative/comments/me9vn0/typescript_use_a_class_or_an_interface_for_api/

https://stackoverflow.com/questions/53531488/why-do-we-need-dtos-and-interfaces-both-in-nestjs

marcepanowyy (changes requested):

backend/src/shared/interfaces/response-objects/BoardResponseObject.ts
Wajktor13: When `BoardResponseObject` is an interface:

image

That's why I changed it to a class. Couldn't find any workaround.

marcepanowyy: When leaving class instead of interface it does not work properly (i think all the fields are optional then)

Wajktor13: So we need both class and interface I guess

Wajktor13: As two separate objects. What do you think? I don't really like it but can't see other way

Wajktor13: Please check this @marcepanowyy : https://github.com/OBELISK-TEAM/OBELISK/compare/90a081ff7922e707b00d646ba3917e5b04a66ad1..4a94e22d6609fb4fdec5779a3e2a754ee7327638

backend/src/modules/boards/boards.controller.ts
```diff
@@ -13,51 +14,113 @@ import { BoardsService } from './boards.service';
 import { CreateBoardDto } from './boards.dto';
 import { User } from '../auth/decorators/users.decorator';
 import { JwtAuthGuard } from '../auth/guards/jwt.auth.guard';
-import { BoardResponseObject } from '../../shared/interfaces/response-objects/BoardResponseObject';
+import { IBoardResponseObject, BoardResponseObject } from '../../shared/interfaces/response-objects/BoardResponseObject';
+import { ApiResponse } from '@nestjs/swagger';
 
 @Controller('boards')
```
marcepanowyy, line 20: no api tags provided (included auth module)
```
@ApiTags('boards')
```

backend/src/modules/boards/boards.controller.ts
```diff
@@ -13,51 +14,113 @@ import { BoardsService } from './boards.service';
...
 export class BoardsController {
   constructor(private readonly boardsService: BoardsService) {}
 
   @Get()
+  @ApiResponse({
```
marcepanowyy, line 25: no api operation provided
```
@ApiOperation({ summary: `Returns a paginated list of all boards' })
```

backend/src/modules/boards/boards.controller.ts
```diff
@@ -13,51 +14,113 @@ import { BoardsService } from './boards.service';
...
+    status: HttpStatus.OK,
+    description: 'returns a paginated list of all boards',
```
marcepanowyy, line 27: if ok then i would change the description:
```
List of all boards retrieved succesfully
```
or sth like that

backend/src/modules/boards/boards.controller.ts
```diff
@@ -13,51 +14,113 @@ import { BoardsService } from './boards.service';
...
+    type: Array<BoardResponseObject>,
```
marcepanowyy, line 28: You could try also with type
```
type: [BoardResponseObject]
```

backend/src/modules/boards/boards.controller.ts
```diff
@@ -13,51 +14,113 @@ import { BoardsService } from './boards.service';
...
+  })
```
marcepanowyy, line 29: no api internal server error response provided
```
@ApiInternalServerErrorResponse({ description: 'Internal server error.' })
```

marcepanowyy (changes requested):

backend/src/shared/interfaces/response-objects/BoardResponseObject.ts
marcepanowyy: hmm i dont like this approach

marcepanowyy: you could try adding @ApiProperty in response objects
```
  @ApiProperty({
    description: 'descr',
    type: [sth],
  })
  prop: sth[];

```

Wajktor13: Yeah, thats the plan for later, but it still doesn't solve the problem.

marcepanowyy: I guess we can stick to the class instead of interface... I see no other solution for that rn :((

PR #67, opened 2024-08-31
feat(slides/slide-objects)[OK-116] Add response object methods

marcepanowyy:


backend/src/modules/slide-objects/slide-objects.service.ts
```diff
@@ -8,82 +8,135 @@ import {
...
   constructor(
     @InjectModel(SlideObject.name)
     private readonly slideObjectModel: Model<SlideObject>,
     private readonly userService: UsersService,
     private readonly slideService: SlidesService,
   ) {}
 
-  async findAll(): Promise<SlideObjectDocument[]> {
-    return this.slideObjectModel.find().exec();
+  async getSlideObjects(
```
Wajktor13, line 23: What is the use case of this method?

backend/src/modules/slide-objects/slide-objects.service.ts
```diff
@@ -8,82 +8,135 @@ import {
...
-    // const user = await this.userService.findOneById(userId);
+  ): Promise<SlideObjectResponseObject> {
+    return this.deleteSlideObjectById(slideObjectId).then(deletedSlideObject =>
+      this.toResponseSlideObject(deletedSlideObject),
+    );
+  }
+
+  ////////////////////////////////////////////////////////////////////////////////////
 
+  async findSlideObjects(
```
Wajktor13, line 93: Again, what is the use case?

Wajktor13 (approved):

backend/src/modules/slide-objects/slide-objects.service.ts
marcepanowyy: To remain consistency - as for the rest services there are methods called this way (its started out when method findUserById was used by auth service and users controller - i decided to divide the method into two - first one returning the whole document and the second one returning only response object. For the future reference, we might need it as well when operating on slide.service

Wajktor13: I don't see the use case for fetching objects from all slides of all boards with pagiantion. I don't think we want to have code that won't be used. However I can live with that so I approve.

PR #70, opened 2024-09-02
refactor(modules)[OK-124] Put in order messy module code

marcepanowyy:


backend/src/modules/auth/auth.controller.ts
```diff
@@ -53,15 +53,15 @@ export class AuthController {
   }
 
   // add @MinimumRole(UserRole.SOMETHING) to secure the route with a minimum role
-  @Get('min-role-secured')
+  @Get('min-role-secured-user')
```
Wajktor13, line 56: Don't we want enums for this?

Wajktor13 (changes requested):

backend/src/modules/auth/auth.controller.ts
marcepanowyy: These endpoints are for testing and developing - it shows how decorators should be used, nothing more.

Wajktor13 (approved):

PR #73, opened 2024-09-08
feat(undo-redo)[OK-95] Add undo-redo board functionality for frontend

Strattoss:
This PR introduces a more advanced undo/redo solution for the Fabric.js board. We no longer need to save entire canvases as JSON objects and store them in memory. Instead, all board actions are encapsulated in individual command objects. The available commands are:

-    `AddCommand`
-    `RemoveCommand`
-    `ModifyCommand`
-    `ComplexCommand`

Eraser paths are handled with `ModifyCommand` as technically eraser adds a path object to the `eraser` property of canvas objects.

I turned off the possibility of scaling and rotating ActiveSelection (if a user selects multiple objects), because handling it and preparing undo/redo commands is a path through ordeal. If, in the future, we find a clever solution and if we have enough time, we can modify the code.

**Important**: From now on, every canvas object must have an `id` property. This allows us to easily identify canvas objects. For now, these ids are assigned by the frontend code, but in the future this responsibility will shift to the server/backend.

**Tests**: @Adam0s007, it would be great if you could run this PR locally and verify if the undo/redo functionality works as intended. If you encounter any issues, please leave a comment or a suggestion. If everything works correctly, please update the label from `test-required` to `tested-ok`.

PS. I hope the code review won't be too much of a hassle 😅


Strattoss:
I need to code 2 more things before we merge: moving multiple objects at the same time and changing the content of a text object
Edit: done

client/src/classes/undo-redo-commands/RemoveCommand.ts
```diff
@@ -0,0 +1,82 @@
+import { FabricObjectIdError } from "@/errors/FabricObjectIdError";
+import { UndoRedoCommand } from "@/interfaces/undo-redo-context";
+import { getItemById } from "@/utils/board/canvasUtils";
+import { toast } from "sonner";
+import { fabric } from "fabric";
+import { addListenersBack, removeListenersTemporarily } from "@/utils/board/undoRedoUtils";
+
+/**
+ * Its purpose is to encompass adding/removing objects DIRECTLY to/from the canvas.
```
adbista: Since this description is identical to the one in AddCommand, it could be misleading. To avoid confusion, consider updating the description to highlight the key differences. (JUST SUGGESTION ;) )

client/src/contexts/UndoRedoContext.tsx
```diff
@@ -1,103 +1,249 @@
...
 
 export const UndoRedoProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
   const {
     state: { canvas },
+    handleStyleChange,
   } = useCanvas();
 
+  // these states are used to memorize the content of recently modified text
+  // so that we can use this old value to prepare a propper ModifiedCommand
+  const [observedTextId, setObservedTextId] = useState<string | undefined>(undefined);
```
adbista: It seems like there are a lot of useStates... Maybe we should consider using useReducer here?

client/src/hooks/board/useMenuActions.ts
```diff
@@ -31,142 +35,201 @@ const getProperties = (color: string, size: number): CanvasActionProperties => (
...
-
 export const useMenuActions = () => {
   const {
     state: { canvas, color, size },
     setCanvasMode,
   } = useCanvas();
-  const { saveState } = useUndoRedo();
+  const { saveCommand } = useUndoRedo();
+
+  const actionHandlers: Record<MenuActions | CanvasMode, CanvasActionHandler> = useMemo(() => {
```
adbista, line 46: Why are we using useMemo here? This is a function, not the result of the function.

client/src/hooks/board/useMenuActions.ts
```diff
@@ -31,142 +35,201 @@ const getProperties = (color: string, size: number): CanvasActionProperties => (
...
+        if (!canvas || !setCanvasMode) {
+          return;
+        }
+        setEraserMode(canvas, size as number, setCanvasMode);
+      },
+      [MenuActions.ADD_LINE]: ({ canvas, properties, setCanvasMode }) => {
+        if (!canvas || !properties || !setCanvasMode) {
+          return;
+        }
+        const addedLine = addLine(canvas, properties);
```
adbista, line 78: It seems like there's a lot of repeated code when adding new objects. I assume you weren't sure how to refactor this elegantly, since the logic for GROUP_SELECTED and the code below it differs?

client/src/contexts/UndoRedoContext.tsx
```diff
@@ -1,103 +1,249 @@
...
+    if (!lastAction) {
+      return;
     }
+
+    lastAction.redo();
+    undoStack.current.push(lastAction);
+    canvas.renderAll();
   }, []);
 
   useEffect(() => {
```
adbista: There’s quite a bit of complex code in this file.  Maybe we could create a custom hook for this useEffect? The logic inside isn’t shared with other components or contexts, and it’s focused on managing listeners.

client/src/contexts/UndoRedoContext.tsx
```diff
@@ -1,103 +1,249 @@
...
+      }
+
+      const oldValues = e.transform?.original; // the original properties values before modification
+      const modifiedObject = e.target;
+
+      if (!oldValues || !modifiedObject) {
+        return;
+      }
+
+      modifiedObject.clone(
```
adbista: Yea, we discussed the `clone(...)` method, and you suggested changing it to something like this:

```js
const objectJSON = fabricObject.toJSON(["id"]);
const clonedObject = Object.clone(objectJSON);
```

If you're able to get that approach working, that would be great! But if it doesn’t work, no worries.

client/src/contexts/UndoRedoContext.tsx
```diff
@@ -1,103 +1,249 @@
...
+    };
+
+    const handleEraserAdded = (e: any) => {
+      if (!e.targets) {
+        return;
+      } // the erased path doesn't collide with enything erasable
+
+      const commands: UndoRedoCommand[] = [];
+      for (const target of e.targets) {
+        target.clone(
```
adbista: same here with clone method

client/src/contexts/UndoRedoContext.tsx
```diff
@@ -1,103 +1,249 @@
...
+        return;
+      }
+
+      const currentText: string = e.target.text;
+      if (currentText === observedTextContent) {
+        // text content hasn't changed
+        return;
+      }
+
+      e.target.clone(
```
adbista: clone method

client/src/components/board/Toolbar/Controls/FontStyleControls.tsx
```diff
@@ -3,18 +3,55 @@ import { Bold, Italic, Underline } from "lucide-react";
...
+      return;
+    }
+
+    const oldValue = modifiedObject[styleKey];
+    const newValue = modifiedObject[styleKey] === valueTrue ? valueFalse : valueTrue;
+
+    setObjectStyle(canvas, modifiedObject, { [styleKey]: newValue });
+    handleStyleChange();
+
+    modifiedObject.clone(
```
adbista: As I mentioned below, if it can be changed, feel free to go ahead with it ;)

client/src/components/board/Toolbar/Controls/ObjectSpecificControls.tsx
```diff
@@ -3,21 +3,57 @@ import ToolbarInput from "@/components/board/Toolbar/ToolbarInput";
...
+
+    // I know it's cheeky, but otherwise we often register the change twice
+    if (newValue === oldValue) {
+      return;
+    }
+
+    setObjectStyle(canvas, modifiedObject, { [key]: newValue });
+    handleStyleChange();
+
+    modifiedObject.clone(
```
adbista: Also here

client/src/classes/undo-redo-commands/ComplexCommand.ts
```diff
@@ -0,0 +1,36 @@
+import { UndoRedoCommand } from "@/interfaces/undo-redo-context";
+
+/**
+ * The class is used in the canvas undo/redo functionality.
+ * It is used to aggregate multiple objects of UndoRedoCommand into one such a command.
+ *
+ * @param commands an array of commands to perform. They should be in the same order as they should be performed during the **redo** method.
+ */
+export class ComplexCommand implements UndoRedoCommand {
```
adbista, line 9: I don't think I got the naming right with 'ComplexCommand.' It's not very clear and could imply many things. Since it's primarily related to the aggregation of multiple objects, maybe 'GroupObjectCommand' would be a better fit? I'm not entirely sure... If you can't think of a better name, feel free to leave it as it is.

client/src/classes/undo-redo-commands/ModifyCommand.ts
```diff
@@ -0,0 +1,95 @@
...
+      throw new FabricObjectIdError(objectBefore);
+    }
+    if (!objectAfter.id) {
+      throw new FabricObjectIdError(objectAfter);
+    }
+    if (objectBefore.id !== objectAfter.id) {
+      throw new Error("Ids of objectBefore and objectAfter are not the same");
+    }
+
+    this._addObjectBefore = new AddCommand(canvas, objectBefore);
```
adbista, line 61: Is there a strong reason for using the 'add' prefix in addObjectBefore and addObjectAfter, and why should AddCommand be included there specifically?

adbista (changes requested):
You did an excellent and complex job here. I don’t know what more to say... respect :) I also added some comments, but they’re not really critical.

client/src/hooks/board/useMenuActions.ts
Strattoss: I guess I was playing around with useCallback and useMemo and I have forgotten to change it back

client/src/hooks/board/useMenuActions.ts
Strattoss: To be honest I haven't even bothered with eliminating code repetition, although I should have 😅
I will look into it

client/src/classes/undo-redo-commands/ComplexCommand.ts
Strattoss: What I meant to convey here by naming this class `ComplexCommand` is that we use the composite design pattern.
https://refactoring.guru/design-patterns/composite
So the sole purpose of `ComplexCommand` is to aggregate multiple commands.

GroupObjectCommand sounds way to fabrick-ish, as if it had something to do with fabric's objects or fabric's groups. Other names I can propose are `CompositeCommand`, `CompoundCommand` or `AggregateCommand`.

client/src/classes/undo-redo-commands/ModifyCommand.ts
Strattoss: Well, the `ModifyCommand` is constructed in such a way that it uses two instances of `AddComand`.
The `addObjectBefore` is used to add the old version of our object (with `redo`) and to remove this old version from the canvas (with `undo`).
The `addObjectAfter` is used to add the new version of our object (with `redo`) and to remove this new version from the canvas (with `undo`). 
So I guess that the most descriptive names of these variables should be `addObjectVersionBeforeModification` and 
`addObjectVersionAfterModification`, but they are a tad mouthful 😆

adbista: Okay, then leave it

client/src/classes/undo-redo-commands/ComplexCommand.ts
adbista: Okay, let's stick with ComplexCommand then. 👍

adbista (commented):
One more thing: while integrating the frontend with the backend, I noticed they use _id for objects. We can also use _id  to ensure consistency.

client/src/hooks/board/useMenuActions.ts
Strattoss: Wait a moment. The `actionHandlers` is not a function, it's an object. If so, `useMemo` seems rather fitting here.

Strattoss:
I've incorporated suggested changes into this PR, so once again I'm asking for a review :-)

adbista (approved):
Nice, thank you :)

PR #75, opened 2024-09-08
test(user-service)[OK-133] Add tests

Wajktor13:


server/package.json
```diff
@@ -29,6 +29,9 @@
       "json",
       "ts"
     ],
+    "moduleNameMapper": {
```
marcepanowyy, line 32: What is this for?

server/src/modules/users/users.service.ts
```diff
@@ -65,7 +65,7 @@ export class UsersService {
   }
 
   async createNewUser(user: CreateUserDto): Promise<UserDocument> {
-    const createdUser = new this.userModel(user);
```
marcepanowyy, line 68: Why was it deleted? The [documentation ](https://docs.nestjs.com/techniques/mongodb) says:

```
import { Model } from 'mongoose';
import { Injectable } from '@nestjs/common';
import { InjectModel } from '@nestjs/mongoose';
import { Cat } from './schemas/cat.schema';
import { CreateCatDto } from './dto/create-cat.dto';

@Injectable()
export class CatsService {
  constructor(@InjectModel(Cat.name) private catModel: Model<Cat>) {}

  async create(createCatDto: CreateCatDto): Promise<Cat> {
    const createdCat = new this.catModel(createCatDto);
    return createdCat.save();
  }

  async findAll(): Promise<Cat[]> {
    return this.catModel.find().exec();
  }
}
```

server/src/modules/users/test/users.service.test.ts
```diff
@@ -0,0 +1,575 @@
...
+
+  const boardId1 = '222';
+
+  const slideObject1 = { height: 20.0, width: 22.2 } as SlideObjectDocument;
+
+  const slideObjectId1 = '2424';
+
+  // Tests
+
+  describe('findUsers', () => {
```
marcepanowyy, line 112: I thought we should test methods that are using helpers, for example
getUsers, getUserById, createUser, updateUser, deleteUser, etc.

marcepanowyy (changes requested):

server/package.json
Wajktor13: To solve this error:

```
 Cannot find module 'src/enums/user.auth.provider' from 'modules/users/test/users.service.test.ts'

       5 | import { HttpException, HttpStatus } from '@nestjs/common';
       6 | import { CreateUserDto, UpdateUserDto } from '../users.dto';
    >  7 | import { UserAuthProvider } from 'src/enums/user.auth.provider';
         | ^
       8 | import { UserRole } from 'src/enums/user.role';
       9 |
      10 | class UserModelMock {

      at Resolver._throwModNotFoundError (../node_modules/jest-resolve/build/resolver.js:427:11)
      at Object.<anonymous> (modules/users/test/users.service.test.ts:7:1)
```

server/src/modules/users/users.service.ts
Wajktor13: Couldn't find a way to mock the constructor. It does the same thing, but this way I can mock it.

Wajktor13: Actually It also does save the doc, so the following `save` is not needed, I'll remove it. [ref](https://mongoosejs.com/docs/api/model.html#Model.create())

server/src/modules/users/test/users.service.test.ts
Wajktor13: I know I've approved it, but I find it very confusing to have `getUsers` and  `findUsers` or `getUserById` and `findUserById`. I'd strongly suggest moving conversion to response objects to controllers.

marcepanowyy: I dont see point testing database operations then

server/src/modules/users/users.service.ts
```diff
@@ -65,8 +65,7 @@ export class UsersService {
...
-    return createdUser.save();
+    return await this.userModel.create(user);
```
marcepanowyy: Please correct for all services for consistency
I think _await_ is unnecessary in methods:

```
async sthMethod(){
    return sthAsync
}
```

marcepanowyy (changes requested):

server/src/modules/users/test/users.service.test.ts
Wajktor13: Well, we are testing business logic behind our service, not database operations. Database operations are mocked.

server/src/modules/users/users.service.ts
Wajktor13: I will change it when adding tests to each of the services. It doesn't hurt us now.

Redundant awaits removed.

marcepanowyy (approved):

PR #78, opened 2024-09-19
fix(default slide)[OK-142] Add slide creation when board is created

marcepanowyy:


server/src/schemas/board.schema.ts
```diff
@@ -70,3 +70,15 @@ export class Board {
 }
 
 export const BoardSchema = SchemaFactory.createForClass(Board);
+
+// // TODO - set everywhere ObjectId type not string
+BoardSchema.pre('save', async function (next) {
+  if (this.slides.length === 0) {
+    const emptySlide = await this.model('Slide').create({
+      version: '5.3.0',
```
Wajktor13, line 78: Why `5.3.0`?

Wajktor13 (changes requested):

server/src/schemas/board.schema.ts
marcepanowyy: the field version must be present in schema, i suppose it indicated the version of fabric js - when creating object on the frontend it has version 5.3.0

Wajktor13 (approved):

PR #80, opened 2024-09-19
feat(Board)[OK-136] Base board API integration & Object creation integration with backend through REST API

adbista:
- Updated the route for retrieving a board from `/board` to `/user-boards/{boardId}/slides/{slideIndex}`.
- Integration for creating boards.
- **UPDATE**: Enhanced the previously created `handleSimpleObjectAdding` function (credit to @Strattoss) to streamline the addition of any object type, making the process more efficient.

**TO BE FIXED:**

- We need to decide how to handle undo/redo functionality with the backend. Should we opt for a simple REST API, or consider a WebSocket-based approach for better real-time interaction?

- Group selection is not functioning correctly because some improvements are required on the backend side.


Strattoss:
I can see you want this PR to fulfil both the OK-136 and OK-152 and that's OK. If I were you, I would delete OK-152 from Jira altogether and change the OK-136 's title and description to match what this PR is about. This way we won't have Jira task marked as done without any PR linked to it.

Strattoss:
If I understand correctly, we are going to wait with merging this PR until backend correctly handles adding rectangles, right?

client/src/app/api/auth/google/route.ts
```diff
@@ -25,6 +25,7 @@ export async function POST(req: NextRequest) {
 
     if (response.ok) {
       const { accessToken } = await response.json();
+      //console.log("accessToken", accessToken);
```
Strattoss, line 28: Did you purposefully leave it here for "future us" or is it a mistake?

client/src/errors/ApiError.ts
Strattoss, line 1: I can see that now we have two different things named `ApiError`: the class in this file and an interface at `@/interfaces/api-error`, so it's rather confusing. May I suggest changing the name of the interface from `ApiError` to `ApiErrorData`? Or something else.

client/src/app/actions/boardActions.ts
```diff
@@ -0,0 +1,30 @@
...
+      }),
+    });
+
+    if (!response.ok) {
+      const errorData: ApiError = await response.json();
+      throw new Error(errorData.message || "Failed to create board");
+    }
+    return await response.json();
+  } catch (error) {
+    console.error("Error creating board:", error);
```
Strattoss: Change to `Error while creating board: `

client/src/app/actions/boardActions.ts
```diff
@@ -0,0 +1,30 @@
...
+        Authorization: `Bearer ${token}`,
+      },
+      body: JSON.stringify({
+        name: `Board-${Date.now()}`,
+      }),
+    });
+
+    if (!response.ok) {
+      const errorData: ApiError = await response.json();
+      throw new Error(errorData.message || "Failed to create board");
```
Strattoss: I'm not convinced if we should throw such a generic error. It would be nice to be able to differentiate between types of errors. What do you think about creating an error class, like `RestApiError` or something like that?
**EDIT**: I can see you've already created `ApiError`. Maybe reuse it here?

client/src/app/actions/slideActions.ts
```diff
@@ -0,0 +1,32 @@
+"use server";
+
+import { getCookie } from "@/utils/authApi";
+import { extractMessagesFromApiError } from "@/lib/toastsUtils";
+import { ApiError } from "@/errors/ApiError";
+
+export async function createCanvasObject(slideId: string, objectData: any): Promise<any> {
+  const accessToken = getCookie("accessToken");
+  //console.log("token", accessToken);
```
Strattoss, line 9: Has this line been left here purposefully or by mistake?

client/src/app/actions/slideActions.ts
```diff
@@ -0,0 +1,32 @@
...
+      body,
+    });
+
+    if (!response.ok) {
+      const reasons = await extractMessagesFromApiError(response);
+      throw new ApiError(reasons);
+    }
+    return await response.json();
+  } catch (error) {
+    console.error("Error creating object:", error);
```
Strattoss: Change to `Error while creating object: `

client/src/interfaces/board-data-response.ts
Strattoss, line 1: What would you say about placing all response interfaces inside a separate directory? You know, inside `client/src/interfaces/**responses**/board-data-response.ts`

client/src/interfaces/permissions-response.ts
Strattoss, line 1: `client/src/interfaces/**responses**/board-data-response.ts`

client/src/interfaces/slide-response.ts
Strattoss, line 1: `client/src/interfaces/**responses**/board-data-response.ts`

client/src/contexts/CanvasContext.tsx
```diff
@@ -12,17 +12,20 @@ import {
...
+  children,
+  boardData,
+}) => {
   const [state, dispatch] = useReducer(canvasReducer, initialState);
   const canvasRef = useRef<HTMLCanvasElement | null>(null);
   const { handleZoom } = useZoom();
   useEffect(() => {
     const newCanvas = initializeCanvas({ current: canvasRef.current });
     dispatch({ type: CanvasReducerAction.SET_CANVAS, canvas: newCanvas });
-
```
Strattoss: Would you be so kind and add this new line back? 😄

Strattoss:
I have found an error. If I try to add a quite big image to the slide, and error occurs: 
obraz
This is what gets printed on the console:
obraz
And an error from nextjs:
obraz
It looks like backend complains that the image is too big and return an error, which blows up our app.

adbista:
> I have found an error. If I try to add a quite big image to the slide, and error occurs: obraz This is what gets printed on the console: obraz And an error from nextjs: obraz It looks like backend complains that the image is too big and return an error, which blows up our app.

Yes, you're right, I also plan to move command-specific functions into a try-catch block. It's important to mention that uploading images from a local computer generates large strings, which makes the images too large. One way to solve this would be for the backend to store those images in buckets and only send us the public URL to those pictures. However, I understand that this won't be implemented for now.

Strattoss (changes requested):
Nicely done! I was worried that integrating frontend with backend would be much messier regarding the `user-boards/[boardId]/slides/[slideIndex]` path, but I turned out pretty well.
I will approve this PR right after you introduce suggested changes

client/src/app/api/auth/google/route.ts
adbista: It’s not necessarily for future use, but it can make things easier, especially when using Postman, as we can quickly access the accessToken by simply copying and pasting it from the console

client/src/app/actions/boardActions.ts
adbista: thanks for this comment/ ApiErrorData will be here and it will be thrown same way as in create object action

client/src/app/actions/slideActions.ts
adbista: as before, maybe we will want to reuse this

client/src/interfaces/board-data-response.ts
adbista: great idea

client/src/interfaces/slide-response.ts
adbista: yes, i will add even ApiErrorResponse here

client/src/contexts/CanvasContext.tsx
adbista: Yes of course ❤️

Strattoss (approved):

PR #82, opened 2024-09-20
feat(ws)[OK-129] Add base impl of ws with auth

marcepanowyy:
Code looks like a crap, needs to be refactored in the future (delivery needed ASAP)
Besides types to pass CI, not much has changed after the demo

The idea behind decorator ws.client.decorator is to obtain user details via @WsUser() or sth like that (similar case to users.decorator) - didnt work for me, thats why its commented. Its not neccessery to implement so I leave it as it is. Maybe there will be time in the future for that.


server/src/gateway/gateway.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { IsNotEmpty, IsString } from 'class-validator';
+
+export class JoinBoardDto {
```
marcepanowyy, line 3: class validators dont work for ws

Wajktor13 (approved):

PR #85, opened 2024-09-21
feat(Slide)[OK-94] Integrate slider with backend

adbista:
### **Key Changes:**

1. **SlideControlsContext**:
   - Centralized logic for:
     - Creating slides.
     - Deleting slides (with a confirmation dialog).
     - Navigating between slides (previous/next, direct pagination, and pagination with hover input on `PaginationEllipsis`).

2. **Refactored Components**:
   - **`SlidePagination`**: Manages pagination logic and UI, now using the context for easier state handling.
   - **`SlideActions`**: Handles slide creation and deletion, utilizing the context for cleaner logic.

3. **Confirmation Modal**:
   - A confirmation dialog ensures that irreversible actions (such as deleting slides) require user confirmation.

4. **New Slide Actions**:
   - Added actions for creating and deleting slides, fully integrated with the context.
   - Added action for revalidating the path to ensure that backend fetches aren’t cached.
### **Testing Instructions:**

Please test this functionality by:
- Creating multiple slides and verifying that the new slides are added correctly.
- Deleting slides and confirming that the deletion works as expected.
- Testing edge cases 

Ensure that all actions work seamlessly, and toast notifications display correctly for both successes and errors.


client/src/app/actions/boardActions.ts
```diff
@@ -7,6 +7,7 @@ import { ApiError } from "@/errors/ApiError";
 
 export async function createBoard(): Promise<BoardDataResponse> {
   const token = getCookie("accessToken");
+  console.log("token", token);
```
Strattoss: Should it be here?

client/src/components/board/SlideControls/DeleteSlideDialog.tsx
Strattoss, line 1: I'm wondering if naming this component `DeleteSlideDialog` is ok. It contains both the delete slide button and the delete slide dialog. What would you say about separating them into two components? Single responsibility principle and such.
This way we could later put all the dialogs into one directory and try to generalize them.

client/src/components/ui/pagination.tsx
Strattoss, line 1: I guess changes in this file are no longer needed, because we don't have `Create slide` button any more.

Strattoss (changes requested):
Just a few changes here and there and it will be ready to merge

Strattoss:
I noticed this thing while testing. Here's my scenario.
1. Create new board. It has 1 slide.
2. Open new browser tab and go to this board. It has 1 slide.
3. In the first tab add new slide. It has 2 slides.
4. Go back to the second tab. Board has 1 slide (as expected - we don't synchronize it).
5. In the second tab create new slide. Board has now 3 slides (as expected - the view has been refreshed).
6. Currently active slide is the second one, although the third one should be active.

Is there something we can do about it? Frankly speaking I'm not sure if we should mind this error at all, because in the future we are going to synchronize slides and slides count with websockets.

What do you think?

adbista:
> I noticed this thing while testing. Here's my scenario.
> 
> 1. Create new board. It has 1 slide.
> 2. Open new browser tab and go to this board. It has 1 slide.
> 3. In the first tab add new slide. It has 2 slides.
> 4. Go back to the second tab. Board has 1 slide (as expected - we don't synchronize it).
> 5. In the second tab create new slide. Board has now 3 slides (as expected - the view has been refreshed).
> 6. Currently active slide is the second one, although the third one should be active.
> 
> Is there something we can do about it? Frankly speaking I'm not sure if we should mind this error at all, because in the future we are going to synchronize slides and slides count with websockets.
> 
> What do you think?

- I think we should wait for websocket integration with slides and discuss that with @Wajktor13  and @marcepanowyy

Strattoss (approved):

PR #87, opened 2024-09-22
feat(ws)[OK-158] Add object action handler

marcepanowyy:


server/src/gateway/gateway.ts
```diff
@@ -5,131 +5,82 @@ import {
...
+import { JoinBoardService } from './providers/join.board.service';
+import { ObjectActionService } from './providers/object.action.service';
+import { ObjectAction } from '../enums/object.action';
+
+// no need to use @UseGuards() decorator here
+// because the WsAuthGuard is already applied in the ConnectionService
+// auth is done on connection, not on message
+// so no need to authenticate every message
+
+@WebSocketGateway(4003, {
```
Wajktor13, line 28: don't we have an env for the ws gateway port?

server/src/gateway/providers/object.action.service.ts
```diff
@@ -0,0 +1,116 @@
...
+    try {
+      switch (action) {
+        case ObjectAction.ADD:
+          return this.handleAddObject(client, data as AddObjectData);
+        case ObjectAction.UPDATE:
+          return this.handleUpdateObject(client, data as UpdateObjectData);
+        case ObjectAction.DELETE:
+          return this.handleDeleteObject(client, data as DeleteObjectData);
+        default:
+          this.emitErrorAndDisconnect(client, 'Invalid action');
```
Wajktor13, line 34: I think that we decided to make dedicated methods for each action, without casting the data.

server/src/modules/slide-objects/slide-objects.service.ts
```diff
@@ -143,6 +147,65 @@ export class SlideObjectsService {
...
+      deletedSlideObject._id.toString(),
+    );
+    await this.slidesService.deleteObjectFromSlide(
+      slideId,
+      deletedSlideObject._id.toString(),
+    );
+    return this.toResponseSlideObject(deletedSlideObject);
+  }
+
+  // END WS methods
```
Wajktor13: Maybe it would be better to make separate service for ws?

Wajktor13 (changes requested):

server/src/modules/slide-objects/slide-objects.service.ts
Wajktor13: For me it's very confusing that for example we have `deleteSlideObject`, `deleteSlideObjectById` and `deleteObject` methods in the same service. It makes you search for the right method every time you want to use it.

server/src/gateway/providers/object.action.service.ts
marcepanowyy: yeah, because before each method validation should be present. Moreover, they need to be wrapped into try-catch block so i think its less code now

server/src/gateway/gateway.ts
marcepanowyy: yeah, but i cant find a proper solution for handling this, 
I found that we can have createGateway function which accepts config service and returns gateway
but this solution is kinda messy

server/src/modules/slide-objects/slide-objects.service.ts
marcepanowyy: it might turn out half of the methods from [bos].service might be useless (but we can keep them until we can realise what we actually need, they might be good for testing)

server/src/modules/slide-objects/ws.objects.service.ts
```diff
@@ -0,0 +1,89 @@
...
+import { SlidesService } from '../slides/slides.service';
+import { SlideObjectResponseObject } from '../../shared/interfaces/response-objects/SlideObjectResponseObject';
+import {
+  ObjectDataProps,
+  ObjectDataPropsWithId,
+} from '../../gateway/gateway.dto';
+import { WsException } from '@nestjs/websockets';
+
+@Injectable()
+export class WsObjectsService {
```
Wajktor13, line 18: `Object` is painfully generic, why not `SlideObject` as in other places in code?

Wajktor13 (changes requested):

server/src/modules/slide-objects/ws.objects.service.ts
marcepanowyy: i mean its not using Object type or sth, the name is shorter and since there are boards, slides and objects it should be clear i guess..

Wajktor13: For me it's very unclear but whatever

Wajktor13 (approved):

PR #89, opened 2024-09-24
feat(Dialog)[OK-143] Add a modal for entering a name for the new board

adbista:
- Added `CreateBoardDialog` component.
- Added global error color support for both light and dark modes, ensuring accessibility (a11y).
- Moved utility files from the `utils` directory to the `lib` directory.
- Moved `user-boards` components into the `user-boards` directory.


client/src/contexts/CreateBoardContext.tsx
Strattoss, line 1: I'm just wondering. Did we really have to create a whole provider for what is basically a simple `createNewBoard` method? It feels like overengineering. Wouldn't a hook be sufficient?

Strattoss (changes requested):
At first I was really surprised. "65 changed files? In such a small commit??". But yeah, you just moved around a few things

client/src/contexts/CreateBoardContext.tsx
adbista: yes. Another button in header needs to have loading state when we click the other button and vice versa

Strattoss (approved):

PR #92, opened 2024-09-30
feat(websockets)[OK-130] Frontend websocket implementation and integration with the backend

Strattoss:
The PR is broad, but the main thing is the integration with the backend regarding websockets and boards/canvases.

1. SocketProvider provides socket instance, which is a core part of this PR. Is used to connect to the backend (and authenticate) and join the chosen board and then slide.
3. Socket utils have predefined methods that are used to propagate necessary changes to the backend.
4. Slide management. Number of slides is correctly managed. All changes (adding/deleting slides) are propagated through the socket to other users.
5. Socket listeners listen for messages from the backend and apply received changes to the canvas (e.g. add/modify/delete objects). 
6. The user is informed when another user joins/leaves the board/slide.
7. Basic local undo/redo command stack. Integration with backend (undo/redo changes are propagated to other users on the board).


client/src/lib/board/socketEmitUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+  JoinBoardData,
+  JoinSlideData,
+  LeaveBoardData,
+  LeaveSlideData,
+  UpdateObjectData,
+} from "@/interfaces/socket/SocketEmitsData";
+import { Socket } from "socket.io-client";
+
+// Utility function to handle optional callbacks
+const emitWithCallback = (socket: Socket, event: string, data: any, callback?: (res: any) => void) => {
```
adbista, line 15: Do we really need this if statement? If callback doesn't exist, callback will be undefined so socket.emit(event,data) should be equal to socket.emit(event, data, callback);

client/src/hooks/board/useCanvasEventListeners.tsx
```diff
@@ -118,37 +146,41 @@ const useCanvasEventHandlers = (
...
-    canvas.on("selection:created", handleMultipleSelections);
-    canvas.on("selection:updated", handleMultipleSelections);
+    const listeners = [
+      { event: "path:created", handler: handlePathCreated },
+      { event: "object:modified", handler: handleObjectModified },
+      { event: "erasing:end", handler: handleEraserAdded },
+      { event: "text:editing:entered", handler: handleTextEditingEntered },
+      { event: "text:editing:exited", handler: handleTextEditingExited },
+      { event: "selection:created", handler: handleMultipleSelections },
+      { event: "selection:updated", handler: handleMultipleSelections },
```
adbista, line 177: We can easily extend functionality by adding more event listeners to ensure smooth interactions with objects on the board, such as updating changes while moving, rotating, or scaling objects. Add the following three listeners
```
{ event: "object:moving", handler: throttledMethod },
{ event: "object:scaling", handler: throttledMethod },
{ event: "object:rotating", handler: throttledMethod }
```

#### Update `package.json` with the following dependencies:
```json
"lodash": "^4.17.21",
"@types/lodash": "^4.17.9" (devDependency)
```
You can create the throttled method by wrapping your modifying function with `throttle(func, milliseconds)`. 
```
import { throttle } from 'lodash';
```
These changes should enhance the user experience when moving, scaling, or rotating objects. Feel free to experiment with them. I believe wrapping `handleObjectModified` with `throttle` should work, but I'm not entirely sure if it will affect the undo/redo functionality. Give it a try—there’s no harm if it doesn’t work. However, it would be great to implement this improvement if possible.

client/src/lib/board/socketEmitUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+  callback?: (res: any) => void
+) => {
+  emitWithCallback(socket, "delete-slide", deleteSlideData, callback);
+};
+
+export const socketEmitJoinSlide = (socket: Socket, joinSlideData: JoinSlideData, callback?: (res: any) => void) => {
+  emitWithCallback(socket, "join-slide", joinSlideData, callback);
+};
+
+export const socketEmitLeaveSlide = (socket: Socket, leaveSlideData: LeaveSlideData, callback?: (res: any) => void) => {
```
adbista, line 49: Is this used somewhere?

adbista (changes requested):
Encountered Bug: The clear canvas action is not being correctly saved to the command stack. Additionally, after refreshing the browser, all previously drawn objects reappear on the canvas instead of being cleared.

client/src/app/user-boards/(realtime-canvas)/[boardId]/slides/[slideIndex]/layout.tsx
```diff
@@ -0,0 +1,78 @@
...
+      toast.error("Socket not available in slides/[slideIndex]/layout.tsx");
+      return;
+    }
+
+    const joinSlideData = { slide: { slideNumber: slideIndexAsNumber } };
+
+    function handleJoinSlide(res: JoinSlideResponse) {
+      logger.log("Joined slide", res);
+      setSlideData(res);
+      setSlideId(res._id);
```
Comment (adbista) [inline client/src/app/user-boards/(realtime-canvas)/[boardId]/slides/[slideIndex]/layout.tsx]: Do we really needthis additional state here? entire res object is saved to slideData, so maybe this action is unnecessary

client/src/lib/board/socketEmitUtils.ts
Strattoss: You are right. I didn't know that js treats "undefined" as "no argument passed".

client/src/hooks/board/useCanvasEventListeners.tsx
Strattoss: I think we should try and implement this functionality (because it's quite useful), but not on this branch. Perhaps it could be a part of another task?
Another thing we should consider is: should we persist these very frequent changes? It feels like the easiest way to kill our database. I think these frequent throttled changes should be broadcasted to other users connected to the board, but shouldn't be persisted.
That's why I think this functionality requires a bit more things to do. Thus, it should be a separate task

client/src/lib/board/socketEmitUtils.ts
Strattoss: No, this emitter is not used anywhere right now, but I thought it would be convenient to have all the possible emit methods defined in one place. Just in case they are needed somewhere. Additionally, this way we can know which messages are valid ones without having to guess or browse through the backend code.

client/src/app/user-boards/(realtime-canvas)/[boardId]/slides/[slideIndex]/layout.tsx
Comment (Strattoss) [reply client/src/app/user-boards/(realtime-canvas)/[boardId]/slides/[slideIndex]/layout.tsx]: Yep, you're right

Strattoss:
> Encountered Bug: The clear canvas action is not being correctly saved to the command stack. Additionally, after refreshing the browser, all previously drawn objects reappear on the canvas instead of being cleared.

I've encountered this bug before and described it on jira https://pawg.atlassian.net/jira/software/c/projects/OK/issues/OK-246

client/src/lib/board/socketEmitUtils.ts
Strattoss: Well, it turns out it's not right XD
If we emit `socket.emit(event, data, callback)` when callback in `undefined`, then in fact we emit an array of values: `[data, undefined]`. So yes, the if statement is needed here.

client/src/hooks/board/useCanvasEventListeners.tsx
adbista: Okay i created task for it

Strattoss:
> Encountered Bug: The clear canvas action is not being correctly saved to the command stack. Additionally, after refreshing the browser, all previously drawn objects reappear on the canvas instead of being cleared.

Also, I've fixed the clear canvas action. Now it should be added to the command stack as usual.

adbista (approved):

Strattoss:
During testing on this branch two issues were discovered:
1. Deleting multiple objects in a short period of time results in faulty persistence
2. Undo / redo functionality needs to be reconsidered and a new approach needs to be thought of.

PR #95, opened 2024-10-01
feat(Boards)[OK-159] Implement user board tables and ui loading for board tables

adbista:
- Mock boards endpoint in BoardsFetcher.ts 
- Create more general Pagination for both Slides and Boards
- Adjust styles for badges


client/src/components/user-boards/TabButtons.tsx
```diff
@@ -1,12 +1,13 @@
 "use client";
-import React, { useState } from "react";
+import React from "react";
 
-interface TabButtonProps {
+interface TabButton {
```
Strattoss: Why have we decided to change this name from `TabButtonProps` to `TabButton`?

client/src/interfaces/responses/user-boards/board-response.ts
```diff
@@ -0,0 +1,10 @@
+export interface BoardResponse {
+  id: string;
+  name: string;
+  modifiedAt: string;
+  createdAt: string;
+  size: number;
+  sharedWith: string[];
+  owner?: string;
+  yourPermission?: string;
```
Strattoss: I think we can safely rename it to `userPermission` or simply `permission`

client/src/components/ExtendedPagination.tsx
Strattoss, line 1: I really hope that generalizing pagination like this doesn't backfire in the future 🤞
Especially if our lists use different systems (like `0, 1, 2, ...` vs `1, 2, 3, ...`)

Strattoss (approved):

client/src/interfaces/responses/user-boards/board-response.ts
adbista: I need to rename that anyway to keep it consistent with the backend

client/src/components/user-boards/TabButtons.tsx
adbista: well... i see TabButtonProps

client/src/components/ExtendedPagination.tsx
adbista: image

As you can see, the numbers are only incremented by 1 when pagination is displayed.

client/src/components/user-boards/TabButtons.tsx
adbista: Oh yeah, this comment is from 3 days ago.

PR #97, opened 2024-10-03
feat(stats)[OK-174] Add base stats module

Wajktor13:


marcepanowyy (commented):
Isn't it better to create stats module that connects others?

Wajktor13:
> Isn't it better to create stats module that connects others?

Yeah, done

marcepanowyy (approved):

PR #102, opened 2024-10-05
feat(boards)[OK-173] Add endpoint for fetching user-related boards

Wajktor13:
Fetching size for board not implemented yet, idk how to do this.

Tested.

image


Wajktor13:
@Adam0s007 @Strattoss FYI: interface changed a little compared to the one in the Jira task

server/src/modules/boards/boards.controller.ts
```diff
@@ -5,32 +5,46 @@ import {
...
 
-  // TODO - implement
   @Get()
   @UseGuards(JwtAuthGuard)
-  getUserBoards(@User('_id') userId: string): string {
-    return `User ${userId} boards IMPLEMENT ME`;
-    // return this.boardsService.getUserBoards(userId);
+  getUserRelatedBoards(
+    @User('_id') userId: string,
+    @Query('boardsGetOption') boardsGetOption: BoardsGetOption,
```
adbista: Just minor thing... since this function is supposed to retrieve user-related boards by default (as its name suggests), perhaps we don't need the ALL_USER_RELATED option in the enum?

Wajktor13: Well, there are 3 options:

- owned by user
- shared for user
- all (2 above combined)

I would stay with this I think

adbista: For me, a query parameter acts as a filter; we provide it when we want specific information. The ALL option implies 'just retrieve everything,' so it's not really necessary in this case. If the function were more general and intended to retrieve all boards from our app, then this query parameter or a variation of it would be essential.

Wajktor13: Okay, I see now. I will correct this.

Wajktor13: Corrected

server/src/shared/interfaces/response-objects/BoardResponseObject.ts
```diff
@@ -12,6 +12,8 @@ export interface BoardResponseObject {
     moderator: string[];
   };
   slides: string[] | Types.ObjectId[];
+  currentUserPermission: string,
```
adbista: Should we use a string here? Since we have specific permissions, it’s not just any text. Perhaps it would be better to use an enum, something like this:

```typescript
export enum BoardPermission {
  NONE = "NONE",
  VIEWER = "VIEWER",
  EDITOR = "EDITOR",
  MODERATOR = "MODERATOR",
  OWNER = "OWNER",
}
```

server/src/shared/interfaces/response-objects/BoardResponseObject.ts
```diff
@@ -12,6 +12,8 @@ export interface BoardResponseObject {
...
+  size: number,
   slide?: SlideResponseObject | string | SuperSlideDocument;
   createdAt?: Date | string | boolean | undefined;
```
adbista: Why do `createdAt` and `updatedAt` have so many types here? Why not just use `Date` or `string`?

adbista (changes requested):
Apologies for not adding these comments earlier; I didn't notice some of the attributes.

server/src/shared/interfaces/response-objects/BoardResponseObject.ts
Wajktor13: It's not possible to change it to str-enums with current design in auth and permissions validation 
image

server/src/shared/interfaces/response-objects/BoardResponseObject.ts
Wajktor13: @marcepanowyy ?

adbista: And same with slide...

marcepanowyy: When we leave createdAt & updatedAt as a Date we get 

```
TS2322: Type string | boolean | undefined is not assignable to type Date | undefined
Type string is not assignable to type Date
BoardResponseObject. ts(16, 3): The expected type comes from property createdAt which is declared here on type
```

This is because in Schema these fields have SchemaTimestampsConfig interface:
  
```
  interface SchemaTimestampsConfig {
    createdAt?: boolean | string;
    updatedAt?: boolean | string;
```
 
You may notice that it might be undefined (?), string, boolean or just a Date...

Btw why asking?

server/src/modules/boards/boards.controller.ts
```diff
@@ -5,32 +5,46 @@ import {
...
   @Get()
   @UseGuards(JwtAuthGuard)
-  getUserBoards(@User('_id') userId: string): string {
-    return `User ${userId} boards IMPLEMENT ME`;
-    // return this.boardsService.getUserBoards(userId);
+  getUserRelatedBoards(
+    @User('_id') userId: string,
+    @Query('activeTab') boardsOwnershipFilterOption: GetBoardsOwnershipFilter,
+    @Query('page') page: number,
+    @Query('perPage') perPage: number,
```
marcepanowyy: I would suggest renaming `perPage` to `limit`

`GET /docs?page=2&limit=10`

server/src/modules/boards/boards.controller.ts
```diff
@@ -5,32 +5,46 @@ import {
...
+      perPage,
+    );
   }
 
   // TODO - check permissions to fetch board
   @Get(':boardId')
-  getBoard(@Param('boardId') boardId: string): Promise<BoardResponseObject> {
-    return this.boardsService.getBoardById(boardId);
+  getBoard(
+    @User('_id') userId: string,
```
marcepanowyy: `@User` wont be defined unless it goes through `@UseGuards(JwtAuthGuard)`

server/src/modules/response/response.service.ts
```diff
@@ -5,28 +5,61 @@ import { SlideResponseObject } from '../../shared/interfaces/response-objects/Sl
...
+    return {
+      boardsPaginated: boardsResponse,
+      currentPage,
+      perPage,
+      totalPages,
     };
   }
 
   toResponseSlide(slide: SuperSlideDocument): SlideResponseObject {
-    const { _id, objects, version } = slide as SuperSlideDocument;
```
marcepanowyy: It should stay `slide as SuperSlideDocument;` - otherwise it throws error when fetching slide... If you didnt modify it on purpose - i think your IDE probably changes it automatically as mine - idk how to fix that for now

server/src/modules/boards/boards.service.ts
```diff
@@ -41,7 +47,45 @@ export class BoardsService {
...
+    return this.res.toResponseBoard(deletedBoard, owner);
+  }
+
+  async getUserRelatedBoards(
+    userId: string,
+    boardsOwnershipFilterOption: BoardsOwnerShipFilterOption,
+    page: number,
+    perPage: number,
+  ): Promise<BoardsPaginatedResponseObject> {
+    const userRelatedBoards: SuperBoardDocument[] = [];
```
marcepanowyy: I think better option would be to fetch from db the exact amount of boards needed. What if user has 100 related boards? Remember - each board keeps slides and slides keep objects as nested docs - it would consume loads of resource. 

You can try with sth like this:

```
@Injectable()
export class MyService {
  constructor(
    @InjectModel(MyDocument.name) private myDocumentModel: Model<MyDocument>,
  ) {}

  async findDocuments(page: number, limit: number): Promise<MyDocument[]> {
    const skip = (page - 1) * limit;
    return await this.myDocumentModel
      .find()
      .skip(skip)
      .limit(limit)
      .exec();
  }
}
```

Keep in mind that you dont need to fetch whole documents, you can specify what fields are needed :) :

 ```
 async findDocuments(page: number, limit: number): Promise<MyDocument[]> {
    const skip = (page - 1) * limit;
    return await this.myDocumentModel
      .find()
      .select('permissions owner')
      .skip(skip)
      .limit(limit)
      .exec();
  }
```

File: server/src/modules/boards/boards.service.ts [hunk 8]
```diff
@@ -107,4 +151,52 @@ export class BoardsService {
...
+        $or: [
+          { 'permissions.viewer': userId },
+          { 'permissions.editor': userId },
+          { 'permissions.moderator': userId },
+        ],
+      })
+      .exec();
+  }
+
+  private sortBoardsByDate(
```
marcepanowyy: sorting could be done through query to db

server/src/modules/boards/boards.service.ts
```diff
@@ -19,9 +21,12 @@ export class BoardsService {
...
   ) {}
 
-  async getBoardById(boardId: string): Promise<BoardResponseObject> {
+  async getBoardById(
+    boardId: string,
+    userId: string,
+  ): Promise<BoardResponseObject> {
     const board = await this.findBoardById(boardId);
-    return this.res.toResponseBoard(board);
+    return this.res.toResponseBoard(board, userId);
```
marcepanowyy: Please, dont modify `toResponseBoard`,
You can add your custom method for handling boards with additional fields, for example `toResponseUserRelatedBoard` or sth like that

It's better to follow backward compatibility rule

marcepanowyy (changes requested):

server/src/shared/interfaces/response-objects/BoardResponseObject.ts
adbista: Can you ensure that the frontend will not receive 'true'/'false' from this?

marcepanowyy: idk why boolean can be a type of createdAt, we are always getting the string/Date so dont worry :)

adbista: Okay, and also could you clarify the type of the slide object? We would like to receive a concrete object, not SlideResponseObject | string | SuperSlideDocument. What exactly will be sent?

marcepanowyy: It depends on the endpoint you are hitting

marcepanowyy: i think we could narrow the type to be just `SlideResponseObject | string`

marcepanowyy: @Adam0s007 its not the scope of PR, we could discuss the response values tomorrow

adbista: As long as this change doesn't affect the backend response (i.e., there are no inconsistencies in mapping to the appropriate interfaces), you're right — this will be outside the scope of this PR

server/src/modules/boards/boards.service.ts
```diff
@@ -107,4 +156,94 @@ export class BoardsService {
...
+
+  async findBoardsSharedForUser(
+    userId: string,
+    page: number,
+    limit: number,
+  ): Promise<SuperBoardDocument[]> {
+    if (limit == 0) return [];
+
+    const skip = (page - 1) * limit;
+    return this.boardModel
```
marcepanowyy, line 165: you could try to define the query and reuse it for countBoardsSharedForUser 
```
  const query = {
    $or: [
      { 'permissions.viewer': userId },
      { 'permissions.editor': userId },
      { 'permissions.moderator': userId },
    ],
  };

  const [total, boards] = await Promise.all([
    this.boardModel.countDocuments(query).exec(),
    this.boardModel
      .find(query)
      .sort([['updatedAt', 'descending']])
      .skip(skip)
      .limit(limit)
      .exec(),
  ]);

  return { total, boards };
```

Maybe it would be even nicer to define counting method that accepts the query and returns the count. Then, no need for `countBoardsSharedForUser` - it would me more generic ;)

server/src/modules/boards/boards.controller.ts
```diff
@@ -5,26 +5,37 @@ import {
...
   @Get()
   @UseGuards(JwtAuthGuard)
-  getUserBoards(@User('_id') userId: string): string {
-    return `User ${userId} boards IMPLEMENT ME`;
-    // return this.boardsService.getUserBoards(userId);
+  getUserRelatedBoards(
+    @User('_id') userId: string,
+    @Query('activeTab') boardsOwnershipFilterOption: GetBoardsOwnershipFilter,
+    @Query('page') page: number,
+    @Query('perPage') limit: number,
```
marcepanowyy: Isnt it better to replace `perPage` with `limit`?

marcepanowyy (changes requested):

server/src/modules/boards/boards.service.ts
marcepanowyy: Also no need to select all the fields from the `board`. We could use `select` to define only needed fields

Wajktor13: I think we need all of them

marcepanowyy (approved):

PR #106, opened 2024-10-06
feat(boards-stats)[OK-175] Add schema and basic logic

Wajktor13:


server/src/schemas/stats/boards.stats.schema.ts
```diff
@@ -0,0 +1,60 @@
+import { Prop, Schema, SchemaFactory } from '@nestjs/mongoose';
+import { Document, Types } from 'mongoose';
+import { StatsUserTimestamp } from 'src/shared/interfaces/StatsUserTimestamp';
+import { Schema as MongooseSchema } from 'mongoose';
+
+@Schema()
```
marcepanowyy: if you add this before your schema, createdAt and updatedAt props will be automatically added 

```
@Schema({
  timestamps: true,
  versionKey: false,
  validateBeforeSave: true,
})
```

server/src/schemas/stats/boards.stats.schema.ts
```diff
@@ -0,0 +1,60 @@
...
+export class BoardsStats extends Document {
+  @Prop({
+    type: Types.ObjectId,
+    ref: 'Board',
+    required: true,
+  })
+  boardId: Types.ObjectId;
```
marcepanowyy: i think `_id` is created by default and it has `ObjectId` type
If it's sth different and i dont understand or you prefer it that way its fine

server/src/shared/interfaces/StatsUserTimestamp.ts
```diff
@@ -0,0 +1,6 @@
+import { Types } from 'mongoose';
+
+export type StatsUserTimestamp = Array<{
```
marcepanowyy, line 3: wouldn't it be better to define the `export class` with `@Prop()` as it is for the rest schemas (like for example `base.object.schema` etc.)?

server/src/modules/stats/boards/boards.stats.service.ts
```diff
@@ -1,4 +1,157 @@
...
 @Injectable()
-export class BoardsStatsService {}
+export class BoardsStatsService {
+  constructor(
+    @InjectModel(BoardsStats.name)
+    private readonly boardStatsModel: Model<BoardsStats>,
+  ) {}
+
+  async initStats(boardId: string): Promise<void> {
+    await this.boardStatsModel.create({ boardId: new Types.ObjectId(boardId) });
```
marcepanowyy: i think there is no need to create ObjectId when using `create` or `find` methods - imo its transformed into type defined in the schema. Correct me if i am wrong

marcepanowyy (approved):

server/src/schemas/stats/boards.stats.schema.ts
Wajktor13: There is no use for this fields.

server/src/schemas/stats/boards.stats.schema.ts
Wajktor13: This is schema for BoardStats, the boardId is a reference to the board.

server/src/shared/interfaces/StatsUserTimestamp.ts
Wajktor13: Not sure what you mean tbh

server/src/modules/stats/boards/boards.stats.service.ts
Wajktor13: fixed

server/src/modules/stats/board/board.stats.service.ts
```diff
@@ -0,0 +1,157 @@
...
+
+@Injectable()
+export class BoardStatsService {
+  constructor(
+    @InjectModel(BoardStats.name)
+    private readonly boardStatsModel: Model<BoardStats>,
+  ) {}
+
+  async initStats(boardId: string): Promise<void> {
+    await this.boardStatsModel.create({ boardId: boardId });
```
marcepanowyy: `await this.boardStatsModel.create({ boardId: boardId });` = `await this.boardStatsModel.create({ boardId });`

its called shorthand property names

https://stackoverflow.com/questions/38948306/what-is-javascript-shorthand-property

image

server/src/modules/stats/board/board.stats.service.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  async removeStats(boardId: string): Promise<void> {
+    await this.boardStatsModel.findOneAndDelete({ boardId: boardId });
+  }
+
+  async updateLastAccessed(boardId: string): Promise<void> {
+    await this.boardStatsModel.updateOne(
+      { boardId: new Types.ObjectId(boardId) },
```
marcepanowyy: I think there is no need for `new Types.ObjectId(boardId)` and it could be shortened to 
`await this.boardStatsModel.updateOne({ boardId })`

marcepanowyy: same situation in other places - its not required - its just my suggestion

marcepanowyy (approved):

PR #107, opened 2024-10-06
feat(BoardDetails)[OK-187] Create proper userboards subdirectories structure to implement board details

adbista:
- Add new project structure for board details

### DIFF
--- a/client/src/components/board-details/BoardDetails.tsx
+++ b/client/src/components/board-details/BoardDetails.tsx
@@ -1,0 +1,8 @@
+// todo: implement board details
+export const BoardDetails = () => {
+  return (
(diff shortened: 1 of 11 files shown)


client/src/components/board-details/BoardDetails.tsx
Strattoss, line 1: If this component is simply a title (h2), maybe name it `BoardDetailsTitle`?

Strattoss (approved):

client/src/components/board-details/BoardDetails.tsx
adbista: But in future it won't be a title

adbista: I will add comment for  this

PR #109, opened 2024-10-06
feat(BoardSidebar)[OK-188] Create BoardSidebar for board details

adbista:


client/src/components/board-details/SidebarButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export const SidebarButton = ({ href, children }: SidebarButtonProps) => {
+  const pathname = usePathname();
+  const isActive = () => {
+    const normalizedPath = pathname.endsWith("/") ? pathname.slice(0, -1) : pathname;
+    const normalizedHref = href.endsWith("/") ? href.slice(0, -1) : href;
+
+    return normalizedPath === normalizedHref;
+  };
```
Strattoss: I'm not sure if the `SidebarButton` component itself should determine if it's active or not. Don't you think that passing down a prop `isActive: boolean` would be a better approach?

Strattoss (approved):

client/src/components/board-details/SidebarButton.tsx
adbista: Okay done

PR #114, opened 2024-10-07
feat(BoardDetails)[OK-209] Implement board details main view

adbista:


client/src/components/board-details/board-information/BoardDetails.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const lastUpdated = board.modifiedAt ? new Date(board.modifiedAt).toLocaleString() : "";
+
+  const totalSizeUsed = board.size;
+  const usedPercentage = board.maxSize ? (totalSizeUsed / board.maxSize) * 100 : 0;
+
+  const collaboratingUsers = [
+    ...(board.permissions.viewer || []),
+    ...(board.permissions.editor || []),
+    ...(board.permissions.moderator || []),
+  ];
```
Strattoss, line 51: Could we please sort this array somehow (for example alphabetically)? This way the user will always see collaborating users in the same, predictable order.

client/src/components/board-details/board-information/BoardNameField.tsx
```diff
@@ -0,0 +1,86 @@
...
+    } else if (e.key === "Escape") {
+      handleCancel();
+    }
+  };
+
+  useEffect(() => {
+    if (board) {
+      setName(board.name);
+    }
+  }, [board]);
```
Strattoss, line 32: I see a missing dependency. `setName` is being used here

client/src/components/loading/BoardDetailsInfoSkeleton.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </header>
+
+      {/* Main Content */}
+      <main className="mt-4 flex flex-col flex-wrap gap-24 p-2 lg:flex-row">
+        {/* Left Column */}
+        <article className="flex flex-col lg:flex-[2]">
+          {/* Board Name Skeleton */}
+          <div className="mb-4 flex flex-col gap-1">
+            <div className="flex items-center">
+              <Skeleton className="mr-1 h-4 w-4" />
```
Strattoss: I think we can get rid of these little skeletons. These were supposed to represent these little icons on the left side of labels, right?

client/src/components/loading/BoardDetailsInfoSkeleton.tsx
```diff
@@ -0,0 +1,99 @@
...
+              <Skeleton className="h-4 w-24" />
+            </div>
+            <Skeleton className="mt-2 h-10 w-full" />
+          </div>
+
+          {/* Owner Skeleton */}
+          <div className="mb-4 flex flex-col gap-1">
+            <div className="flex items-center">
+              <Skeleton className="mr-1 h-4 w-4" />
```
Strattoss: And here

client/src/components/loading/BoardDetailsInfoSkeleton.tsx
```diff
@@ -0,0 +1,99 @@
...
+              <Skeleton className="h-4 w-20" />
+            </div>
+            <Skeleton className="mt-2 h-10 w-full" />
+          </div>
+
+          {/* Creation Date Skeleton */}
+          <div className="mb-4 flex flex-col gap-1">
+            <div className="flex items-center">
+              <Skeleton className="mr-1 h-4 w-4" />
```
Strattoss: And here

client/src/components/loading/BoardDetailsInfoSkeleton.tsx
```diff
@@ -0,0 +1,99 @@
...
+              <Skeleton className="h-4 w-28" />
+            </div>
+            <Skeleton className="mt-2 h-10 w-full" />
+          </div>
+
+          {/* Last Updated Skeleton */}
+          <div className="mb-4 flex flex-col gap-1">
+            <div className="flex items-center">
+              <Skeleton className="mr-1 h-4 w-4" />
```
Strattoss: And here

client/src/components/loading/BoardDetailsInfoSkeleton.tsx
```diff
@@ -0,0 +1,99 @@
...
+              <Skeleton className="h-4 w-24" />
+            </div>
+            <Skeleton className="mt-2 h-10 w-full" />
+          </div>
+
+          {/* Used Space Skeleton */}
+          <div className="mb-4 flex flex-col gap-1">
+            <div className="flex items-center">
+              <Skeleton className="mr-1 h-4 w-4" />
```
Strattoss: And here

client/src/components/loading/BoardDetailsInfoSkeleton.tsx
```diff
@@ -0,0 +1,99 @@
...
+              <Skeleton className="h-4 w-20" />
+            </div>
+            <Skeleton className="mt-2 h-4 w-full" />
+          </div>
+
+          {/* Number of Slides Skeleton */}
+          <div className="mb-4 flex flex-col gap-1">
+            <div className="flex items-center">
+              <Skeleton className="mr-1 h-4 w-4" />
```
Strattoss: And here

Strattoss (changes requested):
Nice job! Just a few small changes and I will be content

Strattoss:
Oh, one more thing I've forgot to mention. Could you modify these two components, which show board size and number of slides? They look like buttons, which isn't that bad yet, but (and that's worse) they behave like buttons. They change color on hover and they seem clickable, because mouse pointer changes shape

https://github.com/user-attachments/assets/a9a8054e-0a24-4e39-818f-245ee381b1a0

client/src/components/board-details/board-information/BoardDetails.tsx
adbista: I will do it, but not here. Sorting is a complex operation that also requires the use of `useMemo`. Since I only need to perform it in the `CollaboratingUsers` component, I will create the `useMemo` with sorting there.

client/src/components/board-details/board-information/BoardNameField.tsx
adbista: I did this on purpose. If you try to add `setName` there, it will cause an infinite loop :PPP. Sometimes, not everything should be added to the dependency array, even if the linter suggests otherwise ;).

Strattoss: Oh, I haven't noticed it :)

Strattoss (approved):

PR #116, opened 2024-10-08
feat(boards)[OK-210] Return permissions for fetching filtered boards

marcepanowyy:
@Adam0s007 example response value of `http://localhost:4000/boards?page=1&limit=2&order=ascending&tab=3`
Please note - `owner` is outside the scope of `permissions`

```
{
    "boards": [
        {
            "_id": "66fd7cd0840df6a72dcdc319",
            "name": "example name",
            "permission": "MODERATOR",
            "permissions": {
                "viewer": [
                    {
                        "_id": "66f921d2367f31f538621c49",
                        "email": "3psyduck281@gmail.com"
                    }
                ],
                "editor": [],
                "moderator": [
                    {
                        "_id": "66f921b5367f31f538621c45",
                        "email": "2psyduck281@gmail.com"
                    }
                ]
            },
            "owner": {
                "_id": "66f84e34a7a2011139c498e5",
                "email": "psyduck281@gmail.com"
            },
            "createdAt": "2024-10-02T17:03:12.309Z",
            "updatedAt": "2024-10-03T21:54:52.380Z"
        },
        {
            "_id": "6702d438a0f9d21d50bf182e",
            "name": "example name",
            "permission": "OWNER",
            "permissions": {
                "viewer": [],
                "editor": [],
                "moderator": []
            },
            "owner": {
                "_id": "66f921b5367f31f538621c45",
                "email": "2psyduck281@gmail.com"
            },
            "createdAt": "2024-10-06T18:17:28.673Z",
            "updatedAt": "2024-10-06T18:17:28.692Z"
        }
    ],
    "page": 1,
    "limit": 2,
    "order": "ascending",
    "total": 6
}
```

### DIFF
--- a/server/src/modules/boards/boards.service.ts
+++ b/server/src/modules/boards/boards.service.ts
@@ -171,7 +171,7 @@ export class BoardsService {
   }
 
   async findUserRelatedBoards(
-    query: FilterQuery<SuperBoard>,
+    query: FilterQuery<SuperBoardDocument>,
     page: number,
     limit: number,
     order: string,
@@ -186,7 +186,9 @@ export class BoardsService {
       .exec();
   }
 
-  async queryCountDocuments(query: FilterQuery<SuperBoard>): Promise<number> {
+  async queryCountDocuments(
+    query: FilterQuery<SuperBoardDocument>,
+  ): Promise<number> {
     return this.boardModel.countDocuments(query).exec();
   }
 }
--- a/server/src/modules/boards/boards.service.ts
+++ b/server/src/modules/boards/boards.service.ts
@@ -54,7 +54,7 @@ export class BoardsService {
     userId: string,
     page: number = 1,
     limit: number = 5,
-    order: SortOrder = 'descending',
+    order: SortOrder = 'desc',
     tab: BoardsFilter = BoardsFilter.ALL,
   ): Promise<PaginatedBoardsResponseObject> {
     const query = this.getFilterQuery(userId, tab);
@@ -186,7 +186,7 @@ export class BoardsService {
     order: SortOrder,
   ): Promise<BoardWithoutSlides[]> {
     const skip = (page - 1) * limit;
-    const sortOrder = order === 'ascending' ? 'ascending' : 'descending';
+    const sortOrder = order === 'asc' ? 1 : -1;
     return this.boardModel
       .find(query)
       .sort([['updatedAt', sortOrder]])
@@ -203,4 +203,4 @@ export class BoardsService {
   }
 }
 
-export type SortOrder = 'ascending' | 'descending';
+export type SortOrder = 'asc' | 'desc';
--- a/server/src/modules/boards/boards.service.ts
+++ b/server/src/modules/boards/boards.service.ts
@@ -11,7 +11,10 @@ import { BoardPermission } from '../../enums/board.permission';
 import { ResponseService } from '../response/response.service';
 import { BoardsFilter } from 'src/enums/boardsFilter';
 import { FilterQueryBuilder } from './filter.query.builder';
-import { SuperBoardWithoutSlides } from '../../shared/interfaces/BoardWithoutSlides';
+import {
+  BoardWithoutSlides,
+  BoardWithPopulatedPermissions,
+} from '../../shared/interfaces/BoardWithoutSlides';
 import { PaginatedBoardsResponse } from '../../shared/interfaces/response-objects/PaginatedUserBoards';
 import { BoardWithSlidesCount } from '../../shared/interfaces/BoardWithSlidesCount';
 import { ClientBoardInfo } from '../../shared/interfaces/ClientBoardInfo';
@@ -53,7 +56,7 @@ export class BoardsService {
     userId: string,
     page: number = 1,
     limit: number = 5,
-    order: string = 'descending',
+    order: SortOrder = 'descending',
     tab: BoardsFilter = BoardsFilter.ALL,
   ): Promise<PaginatedBoardsResponse> {
     const query = this.getFilterQuery(userId, tab);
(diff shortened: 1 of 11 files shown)


server/src/modules/boards/boards.service.ts
```diff
@@ -53,27 +54,35 @@ export class BoardsService {
     userId: string,
     page: number = 1,
     limit: number = 5,
-    order: string = 'descending',
+    order: SortOrder = 'descending',
```
adbista: Could you please shorten the strings to use `desc` and `asc`? Anyway it would be great to generalize the sorting mechanism using parameters. It would be beneficial to follow the convention mentioned in this article:

[Paging, Sorting, Filtering, and Retrieving Specific Fields in Your RESTful API](https://medium.com/@guillaume.viguierjust/paging-sorting-filtering-and-retrieving-specific-fields-in-your-restful-api-a0d289bc574a)

image

server/src/modules/boards/boards.service.ts
```diff
@@ -174,22 +180,27 @@ export class BoardsService {
...
+      .sort([['updatedAt', sortOrder]])
       .skip(skip)
       .limit(limit)
       .select('-slides')
       .exec();
   }
 
-  async queryCountDocuments(query: FilterQuery<SuperBoard>): Promise<number> {
+  async queryCountDocuments(
+    query: FilterQuery<SuperBoardDocument>,
```
adbista, line 200: Please note that we have a filter button planned for future use. Although filtering functionality for boards is not yet implemented, our query filter can be extended to allow filtering based on column properties. For example:

```
?permission=moderator,viewer&name=some_name
```
This query should display only the rows where the user permissions include "moderator" or "viewer" and the board name contains the substring "some_name". While this functionality is not needed immediately, it is important to consider for future development.

server/src/modules/boards/boards.service.ts
```diff
@@ -174,22 +180,27 @@ export class BoardsService {
...
+  ): Promise<number> {
     return this.boardModel.countDocuments(query).exec();
   }
 }
+
+export type SortOrder = 'ascending' | 'descending';
```
adbista: plis, rename to be asc, desc

server/src/shared/interfaces/response-objects/PaginatedUserBoards.ts
```diff
@@ -1,14 +1,24 @@
...
+import { SchemaTimestampsConfig } from 'mongoose';
+import { UserPermission } from '../UserPermission';
 
-export interface PaginatedBoardsResponse {
-  boards: BoardWithoutSlidesAndPermissions[];
+export interface PaginatedBoardsResponseObject {
+  boards: PopulatedBoardResponseObject[];
   page: number;
   limit: number;
   order: string;
```
adbista, line 8: Honestly, I'm not in favor of using the 'order' property just for specifying ascending/descending sorting. While we're not currently using it from the frontend, it would be much clearer to associate the 'order' property with the relevant column(s). I understand that it’s applied to the `updatedAt` field (someone could also assume it is related to createdAt), but if we introduce additional columns in the future, this won't be as intuitive.

For better clarity and flexibility, consider specifying the sort format directly, for example: `sort=some_column,asc` or `sort=-some_column`. This way, it’s easy to extend or modify, such as using `sort=some_column1,some_column2` for multi-column sorting. It’s a more scalable approach.


[Best Practices for Sorting in RESTful APIs](https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#sorting)

image

adbista (changes requested):
For me okay in general, but please consider comments i wrote and shorten ascending, descending

server/src/modules/boards/boards.service.ts
marcepanowyy: I didnt know about this approach, why not mentioned before implementing the endpoint?

server/src/modules/boards/boards.service.ts
marcepanowyy: ok done

server/src/modules/boards/boards.service.ts
marcepanowyy: There wasno info details about adding permission to filter option, how would i know?

server/src/shared/interfaces/response-objects/PaginatedUserBoards.ts
marcepanowyy: I can do that but in a separate ticket

server/src/modules/boards/boards.service.ts
adbista: i didn't go through your code before, i thought that maybe you have in mind we would want to extend some things in future

marcepanowyy: I was told to implement two approaches 
1) fetch boards for which I am an owner
2) fetch boards that are shared for me

I can extend the implementation but in a different ticket. We can discuss the details on the sync in order to avoid similar events.

server/src/shared/interfaces/response-objects/PaginatedUserBoards.ts
marcepanowyy: this approach is quite nice, i will use it for sure

server/src/modules/boards/boards.service.ts
adbista: > There wasno info details about adding permission to filter option, how would i know?

1. That is why it is not needed to be changed in this PR - it is just suggestion to be considered in future
2. I am not saying that specifically permission should have this filter option, but in general those columns

adbista: please do not assume i want something different from which we discussed, i am only pointing out places where changes can be needed in future

adbista (approved):

Wajktor13 (approved):

PR #122, opened 2024-10-08
feat(UserBoard)[OK-200] Integrate endpoint for fetching user-boards with frontend

adbista:
**Why use `useSWR` instead of a simple `fetch` function from server components?**

- Dynamic values that change frequently (such as pagination, sorting, filters (hopefully in the future), and active tabs)
- Two loading states: one with a spinner when the entire page is being rendered, and a second one as a fallback for the page
- Avoids unnecessary code rewrites


client/src/enums/BoardTableColumns.ts
```diff
@@ -5,5 +5,5 @@ export enum BoardTableColumns {
   CREATED_AT = "Created at",
```
marcepanowyy: image

I think a phase `Created at 5 days ago` is incorrect

Shouldnt be either `Created at [specific date]` or just `Created 5 days ago`? Same goes with Modified

marcepanowyy (commented):
image

Is it possible to create 3rd tab with all of the boards?
- I was told to create only those 2 tabs. We discussed this view with @Strattoss, and I don't think it's part of this PR. Also i don't feel we really need this tab. We can discuss this with Gracjan.

marcepanowyy:
I am not familiar with pagination but is this behavior desired?

image

image

Shouldnt the pagination bar be fixed?
- it is kind of tricky to change. Maybe in another PR for fixing this issue
EDIT: okay i probably fixed this issue, but please check it again, okay?

client/src/enums/BoardTableColumns.ts
adbista: Okay done

marcepanowyy (approved):
please update the branch before merging

PR #148, opened 2024-10-20
feat(BoardDetails)[OK-185] Integrate board details view with backend

adbista:
This PR introduces changes in both permissions page and main board details page.

### DIFF
--- a/client/src/components/board-details/board-permissions/BoardPermissions.tsx
+++ b/client/src/components/board-details/board-permissions/BoardPermissions.tsx
@@ -1,6 +1,6 @@
 "use client";
 import { BoardPermissionsUser } from "@/interfaces/board-permissions-user";
-import { useState } from "react";
+import { useCallback, useState } from "react";
 import { BoardPermission } from "@/enums/BoardPermission";
 import { BoardHeader } from "@/components/user-boards/BoardHeader";
 import BoardPermissionsInfoDialog from "@/components/board-details/board-permissions/BoardPermissionsInfoDialog";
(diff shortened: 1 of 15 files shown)


client/src/components/board-details/board-permissions/BoardPermissions.tsx
```diff
@@ -11,38 +11,53 @@ import { Table, TableBody, TableCell, TableRow } from "@/components/ui/table";
...
-      permission: BoardPermission.EDITOR,
-      addedAt: "yesterday 10:42 AM",
-      invitedBy: "Cnon Cnna",
-    },
-  ]);
+import { BoardDetailsResponse } from "@/interfaces/responses/board-details-response";
+import logger from "@/lib/logger";
+export const BoardPermissions = ({ board }: { board: BoardDetailsResponse }) => {
+  const mapPermissions = useCallback((board: BoardDetailsResponse): BoardPermissionsUser[] => {
+    const { viewer, editor, moderator } = board.permissions;
```
Strattoss: `viewer`, `editor` and `moderator` are arrays, right? If so, I would change the name to something more like
```suggestion
    const { viewer: viewers, editor: editors, moderator: moderators } = board.permissions;
```

Strattoss (approved):
Nicee

client/src/components/board-details/board-permissions/BoardPermissions.tsx
adbista: done

adbista: If we want to change that entirely, backend needs to make some changes also

PR #152, opened 2024-10-20
feat(Cursor)[OK-171] Add Cursor tracking on slide

adbista:
I added new cursor functionality for board users on appropriate board. 
Backend:
- Introduce new cursor.action.service.ts
- Emit cursor-remove when we left target (slide, board)

Frontend:
- Add new Cursor component
- Emit our cursor from CanvasContext


server/src/gateway/gateway.ts
```diff
@@ -137,4 +140,17 @@ export class Gateway implements OnGatewayConnection, OnGatewayDisconnect {
   ): Promise<ObjectResponseObject> {
     return this.objectActionService.handleDeleteObject(client, data);
   }
+
+  @SubscribeMessage('cursor-move')
+  @UseGuards(BoardPermissionGuard)
+  @MinimumBoardPermission(BoardPermission.VIEWER)
+  handleCursorMove(
+    client: GwSocketWithTarget,
+    data: CursorData,
```
marcepanowyy: validations wont work for the interface - it needs to be a class as declared in cursor.data

server/src/shared/interfaces/Cursor.ts
```diff
@@ -0,0 +1,14 @@
+export interface CursorData {
```
marcepanowyy: i guess there is no need for the interface

server/src/gateway/providers/cursor.action.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  handleCursorMove(
+    client: GwSocketWithTarget,
+    data: CursorMoveData,
+  ): void {
+    const user  = client.data.user;
+    const { x, y, color } = data;
+    const slideId = client.data.user.targetSlide.slideId;
+
+    if (!slideId) {
```
marcepanowyy, line 17: it cant be such situation - for joined-board users a slideId is always defined

server/src/gateway/providers/cursor.action.service.ts
```diff
@@ -0,0 +1,40 @@
...
+      this.logger.warn(
+        `Client ${user.email} attempted to move cursor without a selected slide.`,
+      );
+      return;
+    }
+
+    client.to(slideId).emit('cursor-move', {
+      userId:user._id,
```
marcepanowyy: wouldnt it be better to return 
```
{
   user: {
            _id: string,
           email: string,
  }
   cursorData: { etc.}

```

server/src/gateway/providers/common.service.ts
```diff
@@ -46,6 +46,7 @@ export class CommonService {
       _id: user._id,
     });
 
+
```
marcepanowyy: no space please

server/src/gateway/dto/cursor.data.ts
```diff
@@ -0,0 +1,21 @@
...
+  @IsNumber()
+  x: number;
+
+  @IsDefined()
+  @IsNumber()
+  y: number;
+
+  @IsDefined()
+  @IsString()
+  color: string;
```
marcepanowyy, line 18: is the color in hex?

server/src/gateway/gateway.ts
```diff
@@ -137,4 +140,17 @@ export class Gateway implements OnGatewayConnection, OnGatewayDisconnect {
...
+  ): void {
+    this.cursorActionService.handleCursorMove(client, data);
+  }
+
```
marcepanowyy: no need for extra spaces

server/src/gateway/providers/cursor.action.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    const slideId = client.data.user.targetSlide.slideId;
+
+    if (!slideId) {
+      this.logger.warn(
+        `Client ${user.email} attempted to move cursor without a selected slide.`,
+      );
+      return;
+    }
+
+    client.to(slideId).emit('cursor-move', {
```
marcepanowyy: maybe its better to emit pattern with past action `cursor-moved` for example?

server/src/gateway/providers/cursor.action.service.ts
```diff
@@ -0,0 +1,40 @@
+// src/gateway/providers/cursor.action.service.ts
```
marcepanowyy: please remove this comment :(

marcepanowyy (changes requested):

server/src/gateway/providers/cursor.action.service.ts
adbista: Actually, i encountered a situation when this if was true, i think to better leave this as it is (i don't know what exactly happened, i just encountered once)

server/src/gateway/dto/cursor.data.ts
adbista: doesn't matter

server/src/gateway/gateway.ts
adbista: Okay, now it is class

server/src/gateway/dto/cursor.data.ts
adbista: I'm not sure what the final color will be; we need to discuss this later. For now, I've only added this color on the client side. We might either go with a random color approach or save a different color property in the database, but I'm not certain yet.

server/src/gateway/providers/cursor.action.service.ts
adbista: Yeah, I think, especially during development, it might be better to notify ourselves when something goes seriously wrong, like when we can't connect to the slide but still emit mouse movement. (That is what happenned before, when one property was wrongly named)

server/src/gateway/dto/cursor.data.ts
Strattoss: May I suggest something? I think it would be nice, if every user had his own, unique colour. This would allow others to easily visually differentiate between users.
What would you say about determining user's colour based on their name / username / nickname? I think Figma does something similar.
Here is the approach I have on my mind: https://medium.com/@femiakt/generate-avatar-or-profile-picture-with-username-initials-lettered-avatar-with-react-eae5d2de5ac8
We could even use this colour to create default avatars and use them here:
obraz

server/src/gateway/providers/cursor.action.service.ts
marcepanowyy: then it should be handled in the guard not in the service - i guess there is some invalid handling

marcepanowyy: and this method should not be responsible for that

server/src/gateway/dto/cursor.data.ts
adbista: Okay, so i propose to create separate task for colors

Strattoss (approved):
What a neat feature!

server/src/gateway/providers/cursor.action.service.ts
adbista: What about object.action.service.ts? There are plenty of these types of checkers.
There is same situation

Strattoss:
I noticed that while zoomed on the canvas, the cursor updates are not correct. The zoomed in canvas is treated as the whole canvas and it doesn't look correct

Now it should be fixed, please check it out again

adbista:
> I noticed that while zoomed on the canvas, the cursor updates are not correct. The zoomed in canvas is treated as the whole canvas and it doesn't look correct
>  

Now it should be fixed, please check it out again

server/src/gateway/providers/cursor.action.service.ts
adbista: done

server/src/gateway/providers/common.service.ts
adbista: done

adbista:
Okay i needed to change ignoreZoom to false, so fabric js has build in solution for this

Strattoss (approved):
Works now

marcepanowyy (approved):

PR #155, opened 2024-10-21
feat(object-stats)[OK-198] Integrate saving

Wajktor13:
For now every object update is an `EDIT_OBJECT` action. In future actions will be determined more precisely, so we will have actions like `MOVE_OBJECT`, `RESIZE_OBJECT` etc.

### DIFF
--- a/server/src/gateway/providers/object.action.service.ts
+++ b/server/src/gateway/providers/object.action.service.ts
@@ -8,11 +8,17 @@ import {
   DeleteObjectData,
   UpdateObjectData,
 } from '../dto/object.data';
+import { ObjectStatsService } from 'src/modules/stats/object/object.stats.service';
+import { SuperObjectDocument } from 'src/mongo/schemas/object/super.object.schema';
+import { Types } from 'mongoose';
 
 @Injectable()
 export class ObjectActionService {
   private readonly logger = new Logger(ObjectActionService.name);
-  constructor(private readonly objectsService: ObjectsService) {}
+  constructor(
+    private readonly objectsService: ObjectsService,
+    private readonly objectStatsService: ObjectStatsService,
+  ) {}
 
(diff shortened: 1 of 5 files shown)


server/src/modules/objects/objects.controller.ts
```diff
@@ -6,10 +6,13 @@ import {
   Body,
   Put,
   Delete,
+  UseGuards,
```
marcepanowyy: objects are not created via REST, i think there is no need to implement this for controllers

server/src/gateway/providers/object.action.service.ts
```diff
@@ -56,6 +57,7 @@ export class ObjectActionService {
       boardId,
       slideId,
       _id,
+      user._id as string,
```
marcepanowyy: I think it would be better to inject in Gateway Module object.stats.service and use it for your purpose in object.action.service. In this approach we are breaking SRP principle by mixing object creation with stats

server/src/modules/objects/objects.module.ts
```diff
@@ -4,11 +4,13 @@ import { ObjectsService } from './objects.service';
 import { ResponseModule } from '../response/response.module';
 
 import { SlidesModule } from '../slides/slides.module';
+import { ObjectStatsModule } from '../stats/object/object.stats.module';
+import { ObjectStatsService } from '../stats/object/object.stats.service';
 
 @Module({
-  imports: [ResponseModule, SlidesModule],
+  imports: [ResponseModule, SlidesModule, ObjectStatsModule],
```
marcepanowyy: As i mentioned before - it would be better to use this module in GW module

marcepanowyy (changes requested):

marcepanowyy (approved):

PR #159, opened 2024-10-22
feat(Auth)[OK-268] Retrieve user info from cookie

adbista:
- Fix caching previous token by replacing router.push with window.location.href. (forcefully reload server components and  then client components). 
- Add "server only" to ensure that some of the functions will only be executed on the server side
- Create function for decoding the token


Strattoss (approved):
Finally, I can easily identify who I am logged in as XD

PR #165, opened 2024-10-28
fix(board)[OK-278] Fix bugs related to join/leave board

adbista:
1. When the user attempted to leave the board by clicking the logo, the connection remained active.

2. When the user tried to revisit the board, the socket from the previous render was retained using useRef

Actions to take:
- Please go to the board (ideally in the first and second browser tabs), click the logo, or the previous page button from the browser, and check if you receive any information about leaving the board in another tab. Then try to return to the same board.
- Next, disconnect the backend and attempt to access the board again.


Strattoss:
The steps you described work as they are supposed to work.

And what a fancy view! I really like it :)
obraz

Strattoss (approved):

adbista:
> The steps you described work as they are supposed to work.
> 
> And what a fancy view! I really like it :) obraz

nice to hear it!

PR #169, opened 2024-10-29
fix(navigation)[OK-283] Fix bugs related to navigation and proper cursor handling

adbista:


client/src/components/board/Cursors/Cursors.tsx
```diff
@@ -36,22 +37,43 @@ const Cursors: React.FC<CursorsProps> = ({ socket, currentUserId }) => {
...
     socket.on("left-slide", handleCursorRemove);
-
-    socket.on("disconnect", () => {
-      setCursorPositions([]);
-    });
+    socket.on("joined-slide", handleCursorJoin);
+    socket.on("disconnect", handleClearCursors);
 
     return () => {
+      console.log("wykonuje sie clear");
```
Strattoss: Delete or change to `logger.info(...)`

client/src/contexts/CanvasContext.tsx
```diff
@@ -126,7 +128,7 @@ export const CanvasProvider: React.FC<CanvasProviderProps> = ({
         const x = pointer.x;
         const y = pointer.y;
 
-        socket?.volatile.emit("cursor-move", { x, y, color: "#aaf" } as any);
+        socket?.volatile.emit("cursor-move", { x, y, color: getColorFromEmail(decodedToken?.email ?? "") } as any);
```
Strattoss: I wonder if it would be worth memorizing the result of the `getColorFromEmail` function with the `useMemo` hook? The `decodedToke?.email` will rarely change

client/src/components/board/Cursors/Cursors.tsx
Strattoss, line 1: Right now we have this:
obraz

What would you say about coloring not the dots of the cursors, but the label instead? The colors would be more pronounced then.

The code would look something like this:
```tsx
return (
      <motion.div
        key={cursor.user._id}
        className={styles.cursorWrapper}
        animate={{ x: clampedX, y: clampedY }}
        transition={{ type: "spring", stiffness: 70, damping: 20 }}
      >
        <div className={styles.cursorLabel} style={{backgroundColor: cursor.cursorData.color}}>{cursor.user.email}</div>
        <div
          className={styles.cursor}
          style={{
            backgroundColor: "#999999",
          }}
        />
      </motion.div>
    );
```

The result:
- colored dots:
obraz
obraz
- hardcoder gray dots:
obraz

We could also introduce some level of transparency (you know, alpha chanel and stuff).

I'm not saying it's better. It's just an alternative option if we want to introduce more colors

Strattoss (changes requested):
I have not encountered any bug (although to be honest I'm not even sure what the bug we are fixing here was about in the first place).
Please, consider my suggestions and you are free to merge

Strattoss:
Oh, one more thing. I think we should change names of directories in the `/components` directory to lowercases. Right now we have `/components/board/Cursors/Cursor.tsx`. I think it would be better to change it into `/components/board/cursors/Cursor.tsx`. And all the other directories, because they look like files (`/components/SlideControl`, `/components/Toolbar`).

adbista:
> I have not encountered any bug (although to be honest I'm not even sure what the bug we are fixing here was about in the first place).
> Please, consider my suggestions and you are free to merge

When i tried to navigate back from canva view i had error message in  bottom-left corner about dehydration. Don't you have the same error?

adbista:
> Oh, one more thing. I think we should change names of directories in the `/components` directory to lowercases. Right now we have `/components/board/Cursors/Cursor.tsx`. I think it would be better to change it into `/components/board/cursors/Cursor.tsx`. And all the other directories, because they look like files (`/components/SlideControl`, `/components/Toolbar`).

ohhhh... yea right

client/src/components/board/Cursors/Cursors.tsx
adbista: what about making cursor darker than label itself? 
image
image
Also i think it would be great to make black/white text in response  to the background color (sometimes background can be too light or too dark)

adbista:
@Strattoss  Could you please test these cursors again and check the changes?

client/src/components/board/Cursors/Cursors.tsx
Strattoss: Good idea, but unfortunately now everything is readable ;(
obraz

Strattoss: Wait, don't we set the same brightness for all the cursors?
obraz
Do you think we could set such values, that a white font would always be suitable?

adbista: Oh... i totally forgot i set this saturation and lightness, then probably there won't be a problem with background at all

Strattoss:
> > I have not encountered any bug (although to be honest I'm not even sure what the bug we are fixing here was about in the first place).
> > Please, consider my suggestions and you are free to merge
> 
> When i tried to navigate back from canva view i had error message in bottom-left corner about dehydration. Don't you have the same error?

No, I haven't seen this error. In fact, I have never ever seen this error, so I don't think I'm a good person to conclude if this error has or hasn't been fixed 🙃
Of course I've tested it again after your changes and it looks good to me. If you want to be sure everything works correctly, you can ask someone else to test it too

client/src/contexts/CanvasContext.tsx
```diff
@@ -37,7 +39,9 @@ export const CanvasProvider: React.FC<CanvasProviderProps> = ({
   const canvasRef = useRef<HTMLCanvasElement | null>(null);
   const { handleZoom } = useZoom();
   const { socket } = useSocket();
-
+  const { decodedToken } = useAuth();
+  const email = decodedToken?.email ?? "";
+  const color = useMemo(() => getColorFromEmail(email), [email]);
```
Strattoss:
```suggestion
  const userColor = useMemo(() => getColorFromEmail(email), [email]);
```
Maybe `userColor`? Or`cursorColor`

Strattoss (changes requested):

Strattoss (approved):
Nicee

adbista:
> > > I have not encountered any bug (although to be honest I'm not even sure what the bug we are fixing here was about in the first place).
> > > Please, consider my suggestions and you are free to merge
> > 
> > 
> > When i tried to navigate back from canva view i had error message in bottom-left corner about dehydration. Don't you have the same error?
> 
> No, I haven't seen this error. In fact, I have never ever seen this error, so I don't think I'm a good person to conclude if this error has or hasn't been fixed 🙃 Of course I've tested it again after your changes and it looks good to me. If you want to be sure everything works correctly, you can ask someone else to test it too



> > > I have not encountered any bug (although to be honest I'm not even sure what the bug we are fixing here was about in the first place).
> > > Please, consider my suggestions and you are free to merge
> > 
> > 
> > When i tried to navigate back from canva view i had error message in bottom-left corner about dehydration. Don't you have the same error?
> 
> No, I haven't seen this error. In fact, I have never ever seen this error, so I don't think I'm a good person to conclude if this error has or hasn't been fixed 🙃 Of course I've tested it again after your changes and it looks good to me. If you want to be sure everything works correctly, you can ask someone else to test it too



> > > I have not encountered any bug (although to be honest I'm not even sure what the bug we are fixing here was about in the first place).
> > > Please, consider my suggestions and you are free to merge
> > 
> > 
> > When i tried to navigate back from canva view i had error message in bottom-left corner about dehydration. Don't you have the same error?
> 
> No, I haven't seen this error. In fact, I have never ever seen this error, so I don't think I'm a good person to conclude if this error has or hasn't been fixed 🙃 Of course I've tested it again after your changes and it looks good to me. If you want to be sure everything works correctly, you can ask someone else to test it too

Okay, I see that when navigating back from the board, everything seems fine. Sorry I wasn’t specific about it—I assumed there would always be an error, but that’s not necessarily the case (though it doesn’t mean everything is working perfectly). Here are the steps to reproduce this hydration error: 1) Create a new board 2) Navigate back.
image
-> Now everything seems to work

PR #171, opened 2024-10-31
feat(envs)[OK-289] Add slide limit per board

marcepanowyy:
It is better to get the value once and have it stored in a private field instead of fetching the value from config sevice over and over again - its not efficient, thats why i modified the boards service


Wajktor13 (dismissed):
When adding new envs, we should also update several places (docker-compose.yml, dockerfiles, github workflows, docs etc.). So either do it or create a task and assign me to it.

marcepanowyy:
@Wajktor13 is it enough? Or should I update other files? 'default' envs were already added to repo settings, please check it out

Wajktor13 (changes requested):
To update after adding new envs:

- `docker-compose.yml`
- client's `Dockerfile` - only if envs needed during build stage
- `.github/docker.yml`
- `docs/INSTALL.md`
- github actions envs

marcepanowyy:
@Wajktor13 

Container (with no client FYI to save up some time)
image

Server logs:
image

working as should

Wajktor13 (approved):

PR #172, opened 2024-10-31
feat(object-stats)[OK-296] Handle delete

Wajktor13:


server/src/modules/boards/boards.service.ts
```diff
@@ -67,6 +69,13 @@ export class BoardsService {
     boardId: string,
   ): Promise<BoardResponseObject> {
     const deletedBoard = await this.deleteBoardById(boardId);
+    for (const slide of deletedBoard.slides) {
```
marcepanowyy: isnt it better to perform deleting logic in the stats service and call the method once implemented?

server/src/gateway/providers/slide.action.service.ts
```diff
@@ -26,6 +30,9 @@ export class SlideActionService {
     const boardId = client.data.user.targetBoard.boardId;
     const slideNumber = data.slide ? data.slide.slideNumber : 1;
     const slide = await this.slidesService.deleteSlide(boardId, slideNumber);
+    for (const obj of slide.objects) {
```
marcepanowyy: isnt it better to perform deleting logic in the stats service and call the method once implemented?

marcepanowyy (approved):

PR #174, opened 2024-10-31
chore(users)[OK-285] Change password and mail requirements

Wajktor13:


server/src/modules/users/users.dto.ts
```diff
@@ -30,17 +30,14 @@ export class CreateUserDto {
   @IsNotEmpty({
     message: 'The password is required',
   })
-  @Length(8, 30, {
+  @Length(8, 64, {
     message:
-      'The password must be at least 8 but not longer than 30 characters',
+      'The password must be at least 8 but not longer than 64 characters',
```
marcepanowyy, line 35: yikes! 64 chars?

marcepanowyy (approved):

server/src/modules/users/users.dto.ts
Wajktor13: image
image

Wajktor13: When you use generated passwords like I do, 30 is nothing.

PR #177, opened 2024-11-01
feat(canvas)[OK-138] Add debouncing technique for actions on board  

adbista:
- Add debounce for object modification listener
- Add debounce in performAction
- Add debounce in toolbar
- create ToolbarContext to remove duplicated code
- Add debounce for undo and redo
- fix Bug with locking text object


client/src/components/board/toolbar-controls/ObjectSpecificControls.tsx
```diff
@@ -1,65 +1,54 @@
...
-    };
-    socketEmitUpdateObject(socket, updateObjectData);
-
-    const objectId: string = modifiedObjectJSON._id;
-    const command = new ModifyCommand(canvas, clonedJSON, modifiedObjectJSON, objectId, handleStyleChange);
-    saveCommand(command);
-  };
+      setObjectStyle(canvas, modifiedObject, { [key]: newValue });
+      handleStyleChange();
+      //logger.log("Toolbar changing object style", key);
```
Strattoss: Delete or uncomment

client/src/contexts/UndoRedoContext.tsx
```diff
@@ -45,6 +47,8 @@ export const UndoRedoProvider: React.FC<{ children: React.ReactNode }> = ({ chil
     lastAction.undo(socket);
     redoStack.current.push(lastAction);
     canvas.renderAll();
+    //toast.dismiss();
+    //toast.message("Undo performed");  for testing purposes
```
Strattoss: For testing purposes :)

client/src/contexts/UndoRedoContext.tsx
```diff
@@ -61,11 +65,21 @@ export const UndoRedoProvider: React.FC<{ children: React.ReactNode }> = ({ chil
     lastAction.redo(socket);
     undoStack.current.push(lastAction);
     canvas.renderAll();
+    //toast.dismiss();
+    //toast.message("Redo performed"); for testing purposes
```
Strattoss: For testing purposes

Strattoss (approved):
I'm truly impressed. You managed to weave the debounce code really elegantly into the existing code.

Strattoss:
I have found a bug. It's easier to observe if debounce times are higher (for example 1 second).
If I select multiple (not one!) objects, move them and deselect them very quickly, the update is not sent to other users.
Here's an example of an update event being sent correctly:

https://github.com/user-attachments/assets/d0f6f431-bfbb-4822-9320-55f88d27bfa0

And here's the situation when the update event is not broadcasted at all to other users:

https://github.com/user-attachments/assets/5f2d6968-78cc-4aab-97a5-986bb28e7392



And I think I know where the problem sits.
We use debounce here:
obraz
But we check, what objects are selected here:
obraz

In the meantime, between the debounced `handleObjectModified` and `handleActiveSelectionModification`, the user can deselect objects or even choose different ones

adbista:
> I have found a bug. It's easier to observe if debounce times are higher (for example 1 second). If I select multiple (not one!) objects, move them and deselect them very quickly, the update is not sent to other users. Here's an example of an update event being sent correctly:
> 
>  correct_update_sent.mp4 
> And here's the situation when the update event is not broadcasted at all to other users:
> 
>  no_update_sent.mp4 
> And I think I know where the problem sits. We use debounce here: obraz But we check, what objects are selected here: obraz
> 
> In the meantime, between the debounced `handleObjectModified` and `handleActiveSelectionModification`, the user can deselect objects or even choose different ones

@Strattoss  Could you please try it again? I think i fixed this issue, it was simple way, i just created two debounced functions: one for handleObjectModified and second for handleActiveSelectionModification.

Adam0s007:
> > I have found a bug. It's easier to observe if debounce times are higher (for example 1 second). If I select multiple (not one!) objects, move them and deselect them very quickly, the update is not sent to other users. Here's an example of an update event being sent correctly:
> > correct_update_sent.mp4
> > And here's the situation when the update event is not broadcasted at all to other users:
> > no_update_sent.mp4
> > And I think I know where the problem sits. We use debounce here: obraz But we check, what objects are selected here: obraz
> > In the meantime, between the debounced `handleObjectModified` and `handleActiveSelectionModification`, the user can deselect objects or even choose different ones
> 
> @Strattoss  Could you please try it again? I think i fixed this issue, it was simple way, i just created two debounced functions: one for handleObjectModified and second for handleActiveSelectionModification.

Strattoss:
> > I have found a bug. It's easier to observe if debounce times are higher (for example 1 second). If I select multiple (not one!) objects, move them and deselect them very quickly, the update is not sent to other users. Here's an example of an update event being sent correctly:
> > correct_update_sent.mp4
> > And here's the situation when the update event is not broadcasted at all to other users:
> > no_update_sent.mp4
> > And I think I know where the problem sits. We use debounce here: obraz But we check, what objects are selected here: obraz
> > In the meantime, between the debounced `handleObjectModified` and `handleActiveSelectionModification`, the user can deselect objects or even choose different ones
> 
> @Strattoss Could you please try it again? I think i fixed this issue, it was simple way, i just created two debounced functions: one for handleObjectModified and second for handleActiveSelectionModification.

It works correctly now

Strattoss:
@Adam0s007 Is there any kind of debouncing for deleting objects? I seems like deleting happens instantaneously

adbista:
> @Adam0s007 Is there any kind of debouncing for deleting objects? I seems like deleting happens instantaneously

Do we need debouncing for this? When we remove objects, they are no longer on the board, so there’s no way we can delete an object multiple times if it’s already removed
image
image

Strattoss:
Oh, yes, you're right

PR #179, opened 2024-11-01
fix(Cursors)[OK-302] Fix bug with cursors position while zoom

adbista:
When the cursors were not moving and I started zooming, their positions did not change. Previously, it worked perfectly. This bug was caused by memoizing the Cursor component. Now, it should work fine.


Strattoss (approved):
Works fine now

PR #180, opened 2024-11-02
feat(share board)[OK-291] Add share button in board view

Strattoss:
I have made slight changes to the code. It's not the prettiest, but works correctly. If you have some other idea, feel free to mention them


adbista (approved):
I think it is okay. For now it can stay as it is.

PR #185, opened 2024-11-03
feat(Permissions)[OK-271] Create viewer view of the board

adbista:
Initially, this branch was created solely to develop the Viewer view of the board. However, after careful consideration and reviewing the user actions table, I realized that we need a more scalable solution to handle cases where specific user actions might change. I’m not entirely sure if this approach is the optimal one, but if we clearly defined which actions apply to which users, we may need to ensure our code can quickly adapt when action permissions change (such as adding, deleting, or updating permissions) 

For example, the Viewer role might currently be restricted from adding or deleting slides, controlling objects, and is only permitted to export files. I could create this view specifically for the Viewer and additional views for roles like Moderator and Owner. However, if permissions change in the future—such as disabling file exports for all roles or restricting slide control to only Moderators—it’s essential that our code can adapt quickly. 

- Adjust our code so that every action in the app is integrated with the corresponding permission.
- Create a `actionPermissionMap` to control each permission.
- Add a hook to resolve the issue with scrolling to the top in the board view.
- Create `generatePermissionFunctions`, which returns an object with dynamically created functions to indicate if a specific action is permitted for a given user.
- Fix bug with BoardDetailsError 

image


client/src/components/board-details/sidebar/SidebarContentClient.tsx
```diff
@@ -0,0 +1,58 @@
...
+      // if user do not have permission to view own stats and view other stats, disable the statistics tab
+      enabled:
+        permissionFunctions.canViewOwnStats(boardData?.permission) ||
+        permissionFunctions.canViewOtherStats(boardData?.permission),
+    },
+  ];
+
+  const normalizePath = (path: string) => (path.endsWith("/") ? path.slice(0, -1) : path);
+
+  const currentPath = normalizePath(pathname);
```
Strattoss: Can you move these lines to the 17th line in this file? I think it would make more sense to have lines 16, 32 and 34 next to each other, because they form a sequence of instructions

client/src/contexts/SocketContext.tsx
```diff
@@ -17,7 +19,7 @@ interface SocketContextProps {
   boardId: string | undefined;
   boardName: string | undefined;
   boardOwner: string | undefined;
-  currentPermission: string | undefined;
+  currentPermission: UserPermissions;
```
Strattoss: I would change the name of this property to `currentPermissions`, because it hold a lot of different permissions now, not just one

client/src/interfaces/user-permissions.ts
```diff
@@ -0,0 +1,12 @@
...
+  canDeleteBoard: boolean;
+  canControlPermissions: boolean;
+  canViewOtherStats: boolean;
+  canViewOwnStats: boolean;
+  canControlSlide: boolean;
+  canControlObject: boolean;
+  canExportBoard: boolean;
+  canViewBoard: boolean;
+  canViewBoardDetails: boolean;
+}
```
Strattoss: What would you say about this?
```suggestion
export interface UserPermissions {
  canEditBoardName: boolean;
  canDeleteBoard: boolean;
  canManageUsersPermissions: boolean;
  canViewOtherStats: boolean;
  canViewOwnStats: boolean;
  canManageSlide: boolean;
  canControlObject: boolean;
  canExportBoard: boolean;
  canViewBoard: boolean;
  canViewBoardDetails: boolean;
}
```

Strattoss (changes requested):
During my review I noticed you have pushed a commit where you changed a lot of things from permission-related to action-related. I am sure this decision was based on some needs, but I think I liked the permission-related names better. Let's talk about it :)

Strattoss:
I have checked (more or less) the code, but I would also like to test this PR in action. I'm afraid I don't have time tonight, so I will do it tomorrow

adbista:
> During my review I noticed you have pushed a commit where you changed a lot of things from permission-related to action-related. I am sure this decision was based on some needs, but I think I liked the permission-related names better. Let's talk about it :)

Permissions are as follows: OWNER, EDITOR, VIEWER, MODERATOR. Actions, however, are more specific. We need to stick to our convention to avoid confusion in naming. When I call the function getPermissions, I expect a list of permissions like OWNER or EDITOR, but that’s not actually what I want — I’m looking for a list of actions

client/src/contexts/SocketContext.tsx
adbista: Well, as I mentioned in another comment 😅😅, maybe we need to distinguish actions from permissions. Otherwise, we'll just end up more confused in the future.
So maybe i will change this to something like: permittedActions or accessRights?

adbista: I commited this already, only because one convention needs to be used. But if you want to discuss this i am open to changes.

client/src/interfaces/user-permissions.ts
adbista: canManageObject also?

Strattoss:
> > During my review I noticed you have pushed a commit where you changed a lot of things from permission-related to action-related. I am sure this decision was based on some needs, but I think I liked the permission-related names better. Let's talk about it :)
> 
> Permissions are as follows: OWNER, EDITOR, VIEWER, MODERATOR. Actions, however, are more specific. We need to stick to our convention to avoid confusion in naming. When I call the function getPermissions, I expect a list of permissions like OWNER or EDITOR, but that’s not actually what I want — I’m looking for a list of actions

Yes, you are totally right. Nonetheless, `actions` sound so generic and non-specific. I feel that if I read the code, I would have troubles understanding that 'actions' are related to what a user has permissions to do in the system.
Do you think that `capabilities` or `privileges` would be better suited here?

client/src/interfaces/user-permissions.ts
Strattoss: Alright

adbista:
> > > During my review I noticed you have pushed a commit where you changed a lot of things from permission-related to action-related. I am sure this decision was based on some needs, but I think I liked the permission-related names better. Let's talk about it :)
> > 
> > 
> > Permissions are as follows: OWNER, EDITOR, VIEWER, MODERATOR. Actions, however, are more specific. We need to stick to our convention to avoid confusion in naming. When I call the function getPermissions, I expect a list of permissions like OWNER or EDITOR, but that’s not actually what I want — I’m looking for a list of actions
> 
> Yes, you are totally right. Nonetheless, `actions` sound so generic and non-specific. I feel that if I read the code, I would have troubles understanding that 'actions' are related to what a user has permissions to do in the system. Do you think that `capabilities` or `privileges` would be better suited here?

Okay done 🫡 Thanks

Strattoss (approved):

Strattoss:
I really like the viewer's view of the board - it's clean and spacious. The only not perfect thing is the footer - it's visible to the user all the time. 
obraz
If you have some idea how to **easily** set the footer below the viewer's view but in such a way that it's still scrollable (just like for a regular user with editing permission), feel free to change this. But if not, don't bother with it! It's really a minor thing

Strattoss:
Thing to add: as an editor I shouldn't be able to see the 'share' button, because I can't really use it:
obraz

adbista:
> Thing to add: as an editor I shouldn't be able to see the 'share' button, because I can't really use it: obraz

Ah yes, i forgot  to add this since this PR was made before yours

adbista:
> I really like the viewer's view of the board - it's clean and spacious. The only not perfect thing is the footer - it's visible to the user all the time. obraz If you have some idea how to **easily** set the footer below the viewer's view but in such a way that it's still scrollable (just like for a regular user with editing permission), feel free to change this. But if not, don't bother with it! It's really a minor thing

I've got one idea, could you try again?

Strattoss (approved):
All clear

PR #186, opened 2024-11-03
refactor(button styles)[OK-272] Adjust user boards row buttons

Strattoss:
Before:
obraz

After: 
obraz


adbista (approved):
Nice small PR

PR #187, opened 2024-11-03
feat(redirect)[OK-292] Redirect to subpage from saved path

adbista:


client/src/lib/urlUtils.ts
```diff
@@ -0,0 +1,27 @@
+import { sanitizeUrl } from "@braintree/sanitize-url";
+import Cookies from "js-cookie";
+
+export const sanitize = (input: string): string => {
+  return sanitizeUrl(input);
+};
```
Strattoss: I'm not sure if this function is needed. We can just use sanitizeUrl directly instead :)

client/src/middleware.ts
```diff
@@ -17,19 +17,26 @@ function isAuthenticated(request: NextRequest): boolean {
...
-  const { pathname } = request.nextUrl;
+  const { pathname, search } = request.nextUrl;
   if (isAuthenticated(request)) {
     if (pathname.startsWith("/auth")) {
       return NextResponse.redirect(new URL("/user-boards", request.url));
     }
   } else {
-    if ((pathname.startsWith("/user-boards") || pathname.startsWith("/board")) && !pathname.startsWith("/auth")) {
-      return NextResponse.redirect(new URL("/auth/login", request.url));
+    if (pathname.startsWith("/user-boards") && !pathname.startsWith("/auth")) {
```
Strattoss: If `pathname` starts with `/user-boards`, can it start with `/auth` at the same time?

Strattoss (changes requested):

client/src/lib/urlUtils.ts
adbista: done

client/src/middleware.ts
adbista: oh yes, done

Strattoss (approved):

PR #198, opened 2024-11-05
feat(modify-permission)[OK-319] Integrate action of updating and deleting permissions with frontend

adbista:
I initially thought this would be easy, but it turned out to be more complicated. The Select component from shadcn/ui behaves in such a way that it always changes its displayed value immediately upon clicking an option. I wanted to create a behavior where, if the backend request fails, the value in the Select field doesn't change. Therefore, I needed to modify the BoardPermissionsSelect component to control when the value updates based on the success of the backend request.


Strattoss:
I haven't noticed it before, but the text in this dialog could be different
obraz
The action is not irreversible, so we can skip that. The rest is ok

Strattoss:
Is this avatar image intended?
obraz

adbista:
> I haven't noticed it before, but the text in this dialog could be different obraz The action is not irreversible, so we can skip that. The rest is ok

Oh yea, i removed it.

adbista:
> Is this avatar image intended? obraz

No... so thanks for noticing this

Strattoss (approved):

PR #206, opened 2024-11-26
feat(board-stats)[OK-339] Add getTimesSpentOnBoardInMs endpoint

Wajktor13:


server/src/modules/users/users.module.ts
```diff
@@ -15,6 +15,6 @@ import { User, UserSchema } from '../../mongo/schemas/user.schema';
   ],
   controllers: [UsersController],
   providers: [UsersService],
-  exports: [UsersService],
+  exports: [UsersService, MongooseModule],
```
marcepanowyy: Why is it exporting mongoose module?

marcepanowyy (changes requested):

server/src/modules/users/users.module.ts
Wajktor13: I get this without:

image

server/src/modules/stats/board/board.stats.module.ts
```diff
@@ -15,9 +17,10 @@ import {
         schema: BoardStatsSchema,
       },
     ]),
+    UsersModule,
   ],
   controllers: [BoardStatsController],
-  providers: [BoardStatsService],
+  providers: [BoardStatsService, UsersService],
```
marcepanowyy: UsersService should not be a provider of this module. Currently it is a provider of UsersModule and it cannot serve as a second provider

server/src/modules/users/users.module.ts
marcepanowyy: no module should export MongooseModule - it should be only imported

marcepanowyy (changes requested):
1) Delete `UsersService` from providers of `BoardStatsModule`
2) Delete `MongooseModule` from exports of `UsersModule`

It will work, however another fix should be made to get rid of `MongooseModule` from `exports` in `BoardStatsModule`

If you have problem with dependencies just draw them - it may simplify debugging

marcepanowyy (approved):

PR #214, opened 2024-12-01
feat(Intervals)[OK-362] Improve aggregation time and include timestamps with zero values

adbista:
The most important aggregation list is here: 
image
maxDays shows when we need to make aggregation
image
determineAggregationRule returns proper object , but instead of enum value, it contain number from enum:
image
image
And last update in StatisticsChart
image
I don't know if it is okay to be in this way, but i just introduced basic option for aggregation.


Wajktor13 (changes requested):
Aggregation interval needs lower values, like 5min, 15min, 0.5h, 1h., maybe up to one week. Bigger values does not make any sense.

adbista:
> Aggregation interval needs lower values, like 5min, 15min, 0.5h, 1h., maybe up to one week. Bigger values does not make any sense.

okay done i think. I also completely removed selection type input. I think this looks better 
image

Wajktor13:
> > Aggregation interval needs lower values, like 5min, 15min, 0.5h, 1h., maybe up to one week. Bigger values does not make any sense.
> 
> okay done i think. I also completely removed selection type input. I think this looks better 
> image
> 

So we don't allow users to specify the aggregation interval?

adbista:
> > > Aggregation interval needs lower values, like 5min, 15min, 0.5h, 1h., maybe up to one week. Bigger values does not make any sense.
> > 
> > 
> > okay done i think. I also completely removed selection type input. I think this looks better
> > image
> 
> So we don't allow users to specify the aggregation interval?

We should? Shouldn't it be automatically adjusted based on start and end date? It doesn't make sense to set aggregation interval to 1 day, when start and enddate are the same. This removed selection was for checking data from previous 1 day, week, 2 weeks, months, and so on. But user can just type for start and end date and thats it

Wajktor13 (approved):

Wajktor13:
> > > > Aggregation interval needs lower values, like 5min, 15min, 0.5h, 1h., maybe up to one week. Bigger values does not make any sense.
> > > 
> > > 
> > > okay done i think. I also completely removed selection type input. I think this looks better
> > > image
> > 
> > 
> > So we don't allow users to specify the aggregation interval?
> 
> We should? Shouldn't it be automatically adjusted based on start and end date? It doesn't make sense to set aggregation interval to 1 day, when start and enddate are the same. This removed selection was for checking data from previous 1 day, week, 2 weeks, months, and so on. But user can just type for start and end date and thats it

IMO we should allow that, but whatever

adbista:
> > > > > Aggregation interval needs lower values, like 5min, 15min, 0.5h, 1h., maybe up to one week. Bigger values does not make any sense.
> > > > 
> > > > 
> > > > okay done i think. I also completely removed selection type input. I think this looks better
> > > > image
> > > 
> > > 
> > > So we don't allow users to specify the aggregation interval?
> > 
> > 
> > We should? Shouldn't it be automatically adjusted based on start and end date? It doesn't make sense to set aggregation interval to 1 day, when start and enddate are the same. This removed selection was for checking data from previous 1 day, week, 2 weeks, months, and so on. But user can just type for start and end date and thats it
> 
> IMO it should, but whatever

@Strattoss  @marcepanowyy  what do you think? client should control this aggregation interval?

marcepanowyy:
Frankly, I did not follow in detail, and I don't know what the aggregation interval is. I believe @Wajktor13 could be a better pick, since he implemented the statistics on the backend side.

Strattoss:
I personally have no strong opinion on that matter. I think we should release a basic version, and possibly consider upgrading it in the future (if ever)

Strattoss (approved):

PR #219, opened 2024-12-07
feat(TimeSpent)[OK-372] Integrate time spent stat endpoints with frontend

adbista:


Strattoss:
@Adam0s007 Right now the pie chart looks like this:
obraz
Would it be possible to add labels like this:
obraz
or a legend like this:
obraz
?
Pretty please 🥺. It would make the chart more visually  appealing

Strattoss:
@Adam0s007 I think that a more useful thing to display in the middle of the chart would be the total amount of time spent on the board by all users (simply the sum of all users' time).
Perhaps something like this?
obraz

client/src/components/common/statistics/StatisticsCard.tsx
Strattoss, line 1: Could you please change the line 20 from
```typescript
{Chart && <main className="mt-4">{Chart}</main>}
```
to
```typescript
{Chart && <main className="my-2">{Chart}</main>}
```
? This way we would achieve a balanced spacing for all situations.

Strattoss (approved):
Just a small change (I now I'm annoying) and we're done

adbista:
> Just a small change (I now I'm annoying) and we're done

I wouldn't say it's annoying; we want our app to be better and better. I know there's a lot of code here, so every minor improvement matters.

PR #222, opened 2024-12-11
feat(board-stats)[OK-338] Add getBoardActionsOverTime endpoint

Wajktor13:


adbista (approved):

server/src/modules/stats/board/board.stats.controller.ts
```diff
@@ -39,4 +39,20 @@ export class BoardStatsController {
   ): Promise<TimeSpentData[]> {
     return this.boardStatsService.getTimesSpentOnBoardInMs(boardId);
   }
+
+  @Get(':boardId/board-actions-over-time')
+  @UseGuards(JwtAuthGuard)
```
marcepanowyy, line 44: Can any authenticated user with a proper board id hit this endpoints? Is it somehow secured besides JwtAuthGuard?

server/src/modules/stats/board/board.stats.service.ts
```diff
@@ -300,4 +300,95 @@ export class BoardStatsService implements OnModuleInit, OnApplicationShutdown {
...
+      aggregationIntervalMinutes * 60 * 1000,
+    );
+
+    return this.convertBoardActionsOverToChartDataArray(
+      boardActionsOverTimeMap,
+    );
+  }
+
+  getBoardActionsOverTimeMap(
+    actionTimeline: [
```
marcepanowyy: Please add interface

marcepanowyy (changes requested):

server/src/modules/stats/board/board.stats.controller.ts
Wajktor13: No, there is a task for securing stats endpoints with permission guard.

server/src/modules/stats/board/board.stats.service.ts
Wajktor13: I changed the parameter to BoardStats and retrieve actionTimeline later

